PR #15, opened 2021-08-16
Update docker push

chinhuang007:
Update docker push script to use the latest commit time to
determine whether to push a new image to Docker Hub.

#### Motivation
No need to push a new image to Docker Hub for the "latest" tag if there is no changes

#### Modifications
Add logic to script to use the latest  commit time to determine whether to publish

#### Result
A new image is pushed to the "latest" tag only if a change is made


pvaneck (commented):
/lgtm
/approve

Issue #14, opened 2021-08-12
Make GitHub Action workflows more selective


pvaneck:
GitHub Action workflows should be more selective on when they run. For example, if just the README or docs are updated, we shouldn't have to run unit tests, although maybe linting is still needed for those. [Path filters](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) can be utilized for these scenarios.

PR #22, opened 2021-08-23
Update workflows to be more selective

pvaneck:
Updated unit-test and docs workflows to not run on a PR with only doc changes. Also made linting a separate action.

Fixes: https://github.com/kserve/modelmesh-serving/issues/14

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,16 +1,14 @@
-name: Unit Test
+name: Lint
 
 on:
   pull_request:
     branches: [main]
     paths-ignore:
-      - 'docs/**'
-      - '**.md'
       - '.github/**'
       - '.tekton/**'
 
 jobs:
-  test:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Lint
+
+on:
+  pull_request:
+    branches: [main]
```
chinhuang007, line 5: Should we also add paths-ignore since lint is for code not docs?

chinhuang007: Not sure if we need to run lint for .github and .tekton since these are excluded in https://github.com/kserve/modelmesh-serving/blob/c77a44f3ed47be7b2d340c0f4928537abb7b142e/.pre-commit-config.yaml#L23

chinhuang007 (approved):
LGTM!

PR #25, opened 2021-08-31
remove changes to the cluster when importing a namespace

xauthulei:
Fix the issue #16 , Thanks

### DIFF
--- a/docs/install/install-script.md
+++ b/docs/install/install-script.md
@@ -87,3 +87,9 @@ To wipe out the ModelMesh Serving CRDs, controller, and built-in runtimes from t
 ```shell
 ./scripts/delete.sh --namespace modelmesh-serving
 ```
+
+(Optional) Delete the specified namespace `modelmesh-serving`
+
+```
+kubectl delete ns modelmesh-serving
+```
--- a/scripts/delete.sh
+++ b/scripts/delete.sh
@@ -67,6 +67,11 @@ if [[ -n $path_to_configs ]]; then
 fi
 
 if [[ -n $namespace ]]; then
+  old_namespace=$(kubectl config  get-contexts $(kubectl config current-context) |tail -1|awk '{ print $5 }')
+  if [[ ! -n $old_namespace ]]; then
+    old_namespace="default"
+  fi
+  echo "current namespace: $old_namespace"
   kubectl config set-context --current --namespace="$namespace"
 fi
 
@@ -79,3 +84,8 @@ kustomize build default | kubectl delete -f - --ignore-not-found=true
 kustomize build runtimes --load-restrictor LoadRestrictionsNone | kubectl delete -f - --ignore-not-found=true
 kubectl delete -f dependencies/quickstart.yaml --ignore-not-found=true
 kubectl delete -f dependencies/fvt.yaml --ignore-not-found=true
+
+# Roll back to previous status
+if [[ -n $namespace ]]; then
+  kubectl config set-context --current --namespace=${old_namespace}
+fi


scripts/delete.sh
```diff
@@ -79,3 +79,9 @@ kustomize build default | kubectl delete -f - --ignore-not-found=true
 kustomize build runtimes --load-restrictor LoadRestrictionsNone | kubectl delete -f - --ignore-not-found=true
 kubectl delete -f dependencies/quickstart.yaml --ignore-not-found=true
 kubectl delete -f dependencies/fvt.yaml --ignore-not-found=true
+
+# Roll back to previous status
+if [[ -n $namespace ]]; then
+  kubectl config set-context --current --namespace="default"
+fi
+kubectl delete ns $namespace
```
chinhuang007, line 87: We don't create namespace in install.sh; therefore probably should not delete namespace automatically in delete.sh. The namespace is created outside of our scripts as instructed here, https://github.com/kserve/modelmesh-serving/blob/main/docs/install/install-script.md. Maybe consider to add a line of instruction in the "Delete the installation" section to manually delete namespace just in case the user wants to keep anything in the namespace after uninstall modelmesh-serving?

chinhuang007: I believe the delete ns command will error out when the optional namespace is not provided.

xauthulei:
@chinhuang007 Would you please help review my new changes, Thanks

docs/install/install-script.md
```diff
@@ -87,3 +87,9 @@ To wipe out the ModelMesh Serving CRDs, controller, and built-in runtimes from t
 ```shell
 ./scripts/delete.sh --namespace modelmesh-serving
 ```
+
+(Optional) Delete the specify namespace `modelmesh-serving`
```
chinhuang007, line 91: Nit, consider "specified" instead of "specify"

scripts/delete.sh
```diff
@@ -79,3 +79,8 @@ kustomize build default | kubectl delete -f - --ignore-not-found=true
 kustomize build runtimes --load-restrictor LoadRestrictionsNone | kubectl delete -f - --ignore-not-found=true
 kubectl delete -f dependencies/quickstart.yaml --ignore-not-found=true
 kubectl delete -f dependencies/fvt.yaml --ignore-not-found=true
+
+# Roll back to previous status
+if [[ -n $namespace ]]; then
+  kubectl config set-context --current --namespace="default"
```
chinhuang007, line 85: Is it possible to capture the previous namespace and reset to the specific name rather than "default"?

docs/install/install-script.md
xauthulei: Good catch, @chinhuang007 , I will update it, Thanks

scripts/delete.sh
xauthulei: @chinhuang007 , I think it is hard to trace in the uninstall stage(or we need caches some things to record this in the install stage), it is best to leave this to use to input. Currently, I think use the `default` is a good choice. Thanks

docs/install/install-script.md
xauthulei: Updated, Thanks

scripts/delete.sh
chinhuang007: I don't think we need to know the namespace used during installation. Can we just capture the current namespace when switching the context in L69-L71 and reset back to that namespace?

xauthulei: Got it， please review this again， Thanks @chinhuang007

scripts/delete.sh
```diff
@@ -67,6 +67,11 @@ if [[ -n $path_to_configs ]]; then
 fi
 
 if [[ -n $namespace ]]; then
+  old_namespace=$(kubectl config  get-contexts $(kubectl config current-context) |tail -1|awk '{ print $5 }')
+  if [[ ! -n $old_namespace ]];then
+	old_namespace="default"
```
chinhuang007, line 72: Everything looks good. Just another nit. Consider to use the same format to be consistent with existing scripts.
1. add space before then
2. use 2 spaces as indentation instead of tab

Thanks!

xauthulei: @chinhuang007 , Fixed, Thanks

pvaneck (approved):
/lgtm
/approve

chinhuang007 (approved):
/lgtm

xauthulei:
@Tomcli , it seems that needing your approve here, Thanks

Tomcli:
The bot cannot add lgtm label to this repo because it's not the owner of the Kserve organization. Not sure did someone changed the bot permission or github api has some new updates.

Tomcli:
I gave the bot admin permission for this repo so it can add lgtm. For other repo, the repo owner needs to promote the kserve-bot as admin from now on.

Tomcli (approved):

Tomcli:
@xauthulei can you force push your commit again? The bot was having some old permission issue to merge the current commit SHA.

xauthulei:
> @xauthulei can you force push your commit again? The bot was having some old permission issue to merge the current commit SHA.

@Tomcli , I force push this again, Thanks

xauthulei:
Thanks again @chinhuang007 , @pvaneck and @Tomcli

Issue #6, opened 2021-08-06
Enable REST support


pvaneck:
Currently only gRPC inference requests are supported with ModelMesh. However, there is a need for REST support in order to smoothen the user experience of interacting with ModelMesh models.

Currently, the idea is to have a transcoder container/service that will transcode the REST v2 protocol JSON into the gRPC format ModelMesh expects (and vice versa). The transcoding mappings will have to be explored, and performance will need to be kept in mind.

pvaneck:
FYI, made a simple prototype: https://github.com/pvaneck/modelmesh-proxy.

This leverages the grpc-gateway project, which there may have been some reservations about, but it seems to work fine for handling rest inference requests.

njhill:
Thanks @pvaneck, this is great. Having looked at grpc-gateway again I realize that it probably could work if we have a custom marshaller. What you have so far though doesn't implement the KFSv2 REST API because it does not map directly to the protobuf one.

i.e. the example in your readme
```
{ "inputs": [{ "name": "predict", "shape": [1, 64], "datatype": "FP32", "contents": { "fp32_contents": [0.0, 0.0, 1.0, ...
```
is not a valid KFSv2 inference request json payload. It would instead be something like
```
{ "inputs": [{ "name": "predict", "shape": [1, 64], "datatype": "FP32", "data": [0.0, 0.0, 1.0, ...
```
See https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md

pvaneck:
@njhill Yea, thanks for pointing that out. My latest iteration does add some custom marshaling logic to enable conformity with the REST v2 spec.  Probably need to revisit the the data array type coercion logic, but just wanted to get it working for now. 

As mentioned in a call last week, right now this transcoder is a separate go program in a container that has its own Deployment and a corresponding Service. I remember you having some thoughts for better integration with modelmesh-serving. Can you expound on what you had in mind?

njhill:
Thanks @pvaneck.

> Can you expound on what you had in mind?

[...]

PR #27, opened 2021-09-14
feat: Add preliminary REST inferencing support via proxy container

pvaneck:
This PR adds preliminary REST support for ModelMesh inferencing. Config options are now available for enabling a REST Proxy container to be deployed as part of a ServingRuntime deployment. 

In the `proxy` directory are the files necessary to build this proxy container. 

Related: #6

### DIFF
--- a/controllers/modelmesh/proxy.go
+++ b/controllers/modelmesh/proxy.go
@@ -1,0 +1,57 @@
+// Copyright 2021 IBM Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package modelmesh
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	restProxyPortEnvVar     = "REST_PROXY_LISTEN_PORT"
+	restProxyGrpcPortEnvVar = "REST_PROXY_GRPC_PORT"
+	restProxyTlsEnvVar      = "REST_PROXY_USE_TLS"
+)
+
+func (m *Deployment) addRESTProxyToDeployment(deployment *appsv1.Deployment) error {
+
+	if m.RESTProxyEnabled {
+		enableTLS := "false"
+		if m.TLSSecretName != "" {
+			enableTLS = "true"
+		}
+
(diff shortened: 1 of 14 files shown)


controllers/modelmesh/proxy.go
```diff
@@ -0,0 +1,57 @@
...
+		if m.TLSSecretName != "" {
+			enableTLS = "true"
+		}
+
+		cspec := corev1.Container{
+			Image: m.RESTProxyImage,
+			Name:  RESTProxyContainer,
+			Env: []corev1.EnvVar{
+				{Name: restProxyPortEnvVar, Value: fmt.Sprintf("%d", m.RESTProxyPort)},
+				{Name: restProxyGrpcPortEnvVar, Value: fmt.Sprintf("%d", m.ServicePort)},
```
njhill, line 42: Use `strconv.Itoa` for these?

controllers/modelmesh/proxy.go
```diff
@@ -0,0 +1,57 @@
...
+				{Name: restProxyTlsEnvVar, Value: enableTLS},
```
njhill, line 43: Could just do `Value: strconv.FormatBool(m.TLSSecretName != "")` here

main.go
```diff
@@ -153,7 +153,8 @@ func main() {
 	conf := cp.GetConfig()
 
 	setupLog.Info("Using adapter", "image", conf.StorageHelperImage.TaggedImage())
-	setupLog.Info("Using model-mesh", "image", conf.ModelMeshImage.TaggedImage())
+	setupLog.Info("Using modelmesh", "image", conf.ModelMeshImage.TaggedImage())
+	setupLog.Info("Using modelmesh REST proxy", "image", conf.RESTProxy.Image.TaggedImage())
```
njhill, line 157: Maybe only log this if the proxy is enabled, and otherwise an info message that it's disabled?

pkg/mmesh/modelmesh_service.go
```diff
@@ -89,6 +94,14 @@ func (mms *MMService) InferenceEndpoint() string {
 	return fmt.Sprintf("%s:%d", mms.Name, mms.Port)
 }
 
+func (mms *MMService) InferenceRESTEndpoint() string {
+	scheme := "http"
```
njhill, line 98: Should probably return `""` if `mms.RESTPort <= 0`? So that we show no http endpoint in the status when the rest proxy is disabled

njhill (commented):
Thanks @pvaneck this looks really good. Just a couple of minor comments.

pkg/mmesh/modelmesh_service.go
pvaneck: Good catch, updated

main.go
pvaneck: Yea, that makes sense. Updated.

controllers/modelmesh/proxy.go
pvaneck: Done

controllers/modelmesh/proxy.go
pvaneck: Done

njhill (approved):

Issue #28, opened 2021-09-16
Reconcile KServe InferenceService CRD


pvaneck:
ModelMesh-Serving needs to be able to reconcile InferenceServices with the `serving.kserve.io/deploymentMode: ModelMesh` annotation. If this annotation does not exist, the resource will be ignored (the KServe controller will handle reconciliation).

Related: https://github.com/kserve/kserve/pull/54

For the first iteration, this will be similar to how a TrainedModel is reconciled in ModelMesh-Serving where TrainedModel fields are mapped internally to PredictorSpec fields as shown [here](https://github.com/kserve/modelmesh-serving/blob/main/apis/kfserving/v1alpha1/trainedmodel_registry.go).

The goal is for an InferneceService YAML like the following to be applied and the corresponding model deployed using ModelMesh:

```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
 name: example-sklearn-mnist-svm
 annotations:
   serving.kserve.io/modelmesh: true
   serving.kserve.io/secret-key: localMinio
Spec:
 predictor:
   sklearn:
     storageUri: s3://sklearn/mnist-svm.joblib
```

Note that this would be the structure with the InferenceService CRD's current state. There is ongoing and upcoming work to introduce a `modelType` field to the InferenceService and also work for storage/credential handling.

PR #34, opened 2021-09-22
feat: Add InferenceService reconciliation

pvaneck:
This allows ModelMesh-Serving to reconcile the KServeInferenceService CRs if the CRD is installed on the cluster. The reconciler will only process the Isvc if a specific `deploymentMode` annotation is set to "ModelMesh".

Closes: #28


apis/serving/v1beta1/inferenceservice_types.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"github.com/kserve/modelmesh-serving/apis/serving/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	SecretKeyAnnotation      = "serving.kserve.io/secret-key"
+	DeploymentModeAnnotation = "serving.kserve.io/deploymentMode"
+	SchemaPathAnnotation     = "serving.kserve.io/schemaPath"
```
njhill, line 29: Style for the names looks inconsistent (snake vs camel)?

apis/serving/v1beta1/inferenceservice_registry.go
```diff
@@ -0,0 +1,138 @@
...
+		},
+	}
+
+	if frameworkSpec != nil && frameworkSpec.RuntimeVersion != nil {
+		p.Spec.Runtime = &v1alpha1.PredictorRuntime{
+			RuntimeRef: &v1alpha1.RuntimeRef{
+				Name: *frameworkSpec.RuntimeVersion,
+			},
+		}
+	}
```
njhill, line 53: Interesting I guess this is good for now, maybe add an explicit comment to say that the string `runtimeVersion` is being interpreted as the name of a ServingRuntime?

apis/serving/v1beta1/inferenceservice_registry.go
```diff
@@ -0,0 +1,138 @@
...
+	return p
+}
+
+func (isvcr InferenceServiceRegistry) Get(ctx context.Context, nname types.NamespacedName) (*v1alpha1.Predictor, error) {
+	inferenceService := &InferenceService{}
+	err := isvcr.Client.Get(ctx, nname, inferenceService)
+
+	if err != nil {
```
njhill, line 61: nit: inline the call to Get?

apis/serving/v1beta1/inferenceservice_registry.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		p.Spec.Path = modelPath
+	}
+	return p, nil
+
+}
+
+func (isvcr InferenceServiceRegistry) UpdateStatus(ctx context.Context, predictor *v1alpha1.Predictor) (bool, error) {
+	inferenceService := &InferenceService{}
+	inferenceService.TypeMeta = predictor.TypeMeta
```
njhill, line 107: Same comment as above ... AFAIK this should be constant for a given CRD.

apis/serving/v1beta1/inferenceservice_registry.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	framework, frameworkSpec := inferenceService.Spec.Predictor.GetPredictorFramework()
+	if frameworkSpec == nil {
+		return nil, errors.New("No valid InferenceService predictor framework found")
+	}
+
+	p := BuildPredictorWithBase(inferenceService, framework, frameworkSpec)
+
+	p.TypeMeta = inferenceService.TypeMeta
```
njhill, line 72: Hmm I'm not sure we want to copy the TypeMeta, rather just set it to `v1.TypeMeta{Kind: "Predictor", APIVersion: "serving.kserve.io/v1"}` (though for practical purposes currently I don't think it makes a difference).

apis/serving/v1beta1/inferenceservice_registry.go
```diff
@@ -0,0 +1,138 @@
...
+	p.ObjectMeta = inferenceService.ObjectMeta
+	p.Status = inferenceService.Status.PredictorStatus
+
+	if p.Status.ActiveModelState == "" {
+		p.Status.ActiveModelState = v1alpha1.Pending
+	}
+	storageUri := *frameworkSpec.StorageURI
+	if !strings.HasPrefix(storageUri, "s3://") {
+		p.Spec.Storage = &v1alpha1.Storage{
+			S3: nil,
```
njhill, line 82: This is so that it will intentionally fail presumably?

njhill (commented):
Thanks @pvaneck this looks great.

I'm actually preparing some changes to refactor some of the predictor source abstraction stuff, which will mean that this needs some tweaks (should be small and a simplification though), so probably best to hold of merging this one until that's in.

I assume we'll rip out `TrainedModel` support at some point after, I agree it's nicer to do that as a separate PR.

controllers/servingruntime_controller.go
```diff
@@ -321,6 +323,24 @@ func (r *ServingRuntimeReconciler) runtimeHasPredictors(ctx context.Context, rt
...
+		inferenceServices := &servingv1beta1.InferenceServiceList{}
+		err = r.Client.List(ctx, inferenceServices, client.InNamespace(r.DeploymentNamespace))
+		if err != nil {
+			return false, err
+		}
+		for i := range inferenceServices.Items {
+			isvc := inferenceServices.Items[i]
+			if isvc.ObjectMeta.Annotations[servingv1beta1.DeploymentModeAnnotation] == servingv1beta1.MMDeploymentModeVal {
+				framework, frameworkSpec := isvc.Spec.Predictor.GetPredictorFramework()
+				p := servingv1beta1.BuildPredictorWithBase(&isvc, framework, frameworkSpec)
```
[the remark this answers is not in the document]
pvaneck: Yea, I updated the annotation check to be in the `BuildPredictorWithBase` function. Noticed that the predictor controller watcher was still trying to reconcile `InferenceServices` without the annotation, so made additional adjustments to fix that.

apis/serving/v1beta1/inferenceservice_registry.go
pvaneck: I guess my intent here was to pass the TypeMeta through the Predictor so that `UpdateStatus` will get the original InferenceService TypeMeta, however, after more fiddling, seems like I can just completely remove these TypeMeta lines, and everything still seems to work fine. Perhaps fine to just remove?

pvaneck (commented):
Thanks for the review, @njhill. And right, the TrainedModel CRD which is still using the `serving.kubeflow.org` group will be removed soon.

apis/serving/v1beta1/inferenceservice_types.go
```diff
@@ -0,0 +1,125 @@
...
+	Type string `json:"type"`
+
+	Status corev1.ConditionStatus `json:"status"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// InferenceService is the Schema for the inferenceservices API
+type InferenceService struct {
```
yuzisun, line 65: Is there any reason why we can't use the type from kserve and import the go module?

apis/serving/v1beta1/inferenceservice_registry.go
yuzisun: Let's see the difference between KServe and ModelMesh here:
- ModelFramework/Version: this is KServe's framework list on the `InferenceService` predictor spec, in ModelMesh this is `ModelType` on predictor
- ServingRuntime/Version:  the `runtimeVersion` currently actually is meant for the ServingRuntime's image version, so I think we should add the `servingRuntime` field on the `InferenceService` predictor spec

apis/serving/v1beta1/inferenceservice_types.go
pvaneck: I recall trying this before and running into some go dependency issues. This was just easier for me at this current stage.

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if p.Status.ActiveModelState == "" {
+		p.Status.ActiveModelState = v1alpha1.Pending
+	}
+
+	_, frameworkSpec := inferenceService.Spec.Predictor.GetPredictorFramework()
+	storageUri := *frameworkSpec.StorageURI
+	if !strings.HasPrefix(storageUri, "s3://") {
+		p.Spec.Storage = &v1alpha1.Storage{
+			S3: nil,
```
njhill, line 66: nit: no need for this, will be nil by default

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -0,0 +1,140 @@
...
+	return p, nil
+
+}
+
+func (isvcr InferenceServiceRegistry) Find(ctx context.Context, namespace string,
+	predicate func(*v1alpha1.Predictor) bool) (bool, error) {
+
+	list := &v1beta1.InferenceServiceList{}
+	err := isvcr.Client.List(ctx, list, client.InNamespace(namespace))
+	if err != nil {
```
njhill, line 94: nit: could inline this

njhill:
> The current logic is to just call handlePredictorNotFound when the annotation is not found in the the isvc. Then when `DeleteVModel` is called, the following is returned: `rpc error: code = Unavailable desc = last resolver error: produced zero addresses` which prompts the requeue of the request and the loop. What is the intent behind the requeue here?

Ah, this is not directly related to the `InferenceService` handling but kind of more exposed/exacerbated by it. It is attempting to call "unregister" in model-mesh (which is actually correct behaviour - imagine if the annotation had previously been there but then deleted), but because there are no other multi-model predictors there are no model-mesh pods running and so it retries until some model-mesh pod comes up. We have some rework planned which would address this, but I will think more about how we can avoid the indefinite retries in the meantime.

controllers/predictor_controller.go
```diff
@@ -102,6 +107,11 @@ func (pr *PredictorReconciler) ReconcilePredictor(ctx context.Context, nname typ
 			nname.Name, err)
 	}
 
+	// Ignore resource if nil predictor is retrieved.
+	if predictor == nil {
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
pvaneck: Hey, the logic was reworked a bit, and now there should be some "Ignoring InferenceService" message for cases of Non-MM InferenceServices being deployed.

controllers/predictor_controller.go
```diff
@@ -257,6 +263,11 @@ func (pr *PredictorReconciler) handlePredictorNotFound(ctx context.Context,
 	_, err := mmc.DeleteVModel(deleteCtx, &mmeshapi.DeleteVModelRequest{VModelId: name.Name, Owner: sourceId})
 	if err != nil {
 		if isNoAddresses(err) {
+			// Work-around to prevent Non-MM InferenceService indefinite reconcile loop.
```
njhill, line 266:
```suggestion
			// Work-around to prevent Non-MM InferenceService indefinite reconcile loop when there are no model-mesh pods running.
```

controllers/predictor_controller.go
```diff
@@ -257,6 +263,11 @@ func (pr *PredictorReconciler) handlePredictorNotFound(ctx context.Context,
...
+			if sourceId == InferenceServiceCRSourceId {
+				pr.Log.Info("Ignoring InferenceService")
```
njhill, line 268: I'm not sure we want to log this every time since there will be a lot of such messages in the case where there are no model-mesh pods and lots of other non-MM `InferenceService` activity.

PR #42, opened 2021-10-06
docs: Add info about using REST proxy

pvaneck:
#### Motivation

Preliminary REST proxy support was added. Some information on usage is needed.

#### Modifications

Added docs about performing REST inference.

#### Result

More usage information for users.


README.md
```diff
@@ -34,6 +34,10 @@ ModelMesh Serving provides out-of-the-box integration with the following model s
 
 `ServingRuntime` custom resources can be used to add support for other existing or custom-built model servers, see the docs on [implementing a custom Serving Runtime](./docs/runtimes/custom_runtimes.md)
 
+### Supplementary
+
```
chinhuang007, line 38: Is it intentional to make "Supplementary" a level higher (with ###) than others like Core Component above and Libraries below? If so, does it mean "Libraries" also Supplementary?

PR #44, opened 2021-10-07
Update config for release v0.7.0

chinhuang007:
Update config for release v0.7.0 to use specific tags
* modelmesh: v0.7.0
* modelmesh-runtime-adapter: v0.7.0
* rest-proxy: v0.1.0


chinhuang007:
Right, this feels like a chicken and egg situation. I will change it and we should create the rc1 release/tag quickly.

Issue #53, opened 2021-10-15
Error for golangci-lint when using go 1.17


pugangxa:
**Describe the bug**
- Error for golangci-lint when using go 1.17
- Lint error when upgrade golangci-lint version

**To Reproduce**

```
➜  modelmesh-serving git:(remove-trainedmodel) ✗ go version
go version go1.17.1 darwin/amd64
```

When using original v1.32.0 of golangci-lint 
```
➜  modelmesh-serving git:(remove-trainedmodel) ✗ pre-commit run --all-files
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 2

fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb01dfacedebac1e pc=0x7fff20425c9e]

runtime stack:
runtime: unexpected return pc for runtime.sigpanic called from 0x7fff20425c9e
stack: frame={sp:0x7ffeefbff218, fp:0x7ffeefbff268} stack=[0x7ffeefb802b8,0x7ffeefbff320)
0x00007ffeefbff118:  0x01007ffeefbff138  0x0000000000000004
0x00007ffeefbff128:  0x000000000000001f  0x00007fff20425c9e
0x00007ffeefbff138:  0x0b01dfacedebac1e  0x0000000000000001
0x00007ffeefbff148:  0x0000000004038831 <runtime.throw+0x0000000000000071>  0x00007ffeefbff1e8
0x00007ffeefbff158:  0x00000000049f4939  0x00007ffeefbff1a0
0x00007ffeefbff168:  0x0000000004038ae8 <runtime.fatalthrow.func1+0x0000000000000048>  0x00000000051fbd40
0x00007ffeefbff178:  0x0000000000000001  0x0000000000000001
0x00007ffeefbff188:  0x00007ffeefbff1e8  0x0000000004038831 <runtime.throw+0x0000000000000071>
0x00007ffeefbff198:  0x00000000051fbd40  0x00007ffeefbff1d8
0x00007ffeefbff1a8:  0x0000000004038a70 <runtime.fatalthrow+0x0000000000000050>  0x00007ffeefbff1b8
0x00007ffeefbff1b8:  0x0000000004038aa0 <runtime.fatalthrow.func1+0x0000000000000000>  0x00000000051fbd40
0x00007ffeefbff1c8:  0x0000000004038831 <runtime.throw+0x0000000000000071>  0x00007ffeefbff1e8
......
```

[...]

PR #54, opened 2021-10-15
Remove trained model

### LINKED ISSUE #51: Remove TrainedModel CR reconciliation
With the TrainedModel CRD being deprecated in favor of using the InferenceService CRD for both single model and multi model deployments, we can go ahead and remove support for TrainedModel reconciliation. Our current support is still using the old `serving.kubeflow.org/v1alpha1`  APIGroupVersion anyway.

pugangxa:
#### Motivation
Fix #51 
#### Modifications
Removed all the related stuff about TrainedModel
#### Result
Build manager and run UT successfully.

I bypassed pre-commit stage due to issue #53


main.go
```diff
@@ -311,7 +306,7 @@ func main() {
 		dispatchers := make([]func(), 0, len(sources))
 		for _, s := range sources {
 			sid := s.GetSourceId()
-			if sid == "" || sid == controllers.PredictorCRSourceId || sid == controllers.TrainedModelCRSourceId {
```
pvaneck, line 314: Ah, just noticed that we are missing the check for controllers.InferenceServiceCRSourceId here. Can you just replace
`controllers.TrainedModelCRSourceID` here with `controllers.InferenceServiceCRSourceId` ?

pvaneck (commented):
Thanks @pugangxa for picking this up. Looks good! Can you try regenerating the deepcopy files using the developer image outlined in my comment [here](https://github.com/kserve/modelmesh-serving/issues/53#issuecomment-944781460). Trying to keep things in line with the various versions in the dev image.

main.go
pugangxa: Sure, changed according to comments.

PR #56, opened 2021-10-16
feat: storage phase 1 for inference service reconciler

Tomcli:
#### Motivation
rebase #32 to the new inference service reconciler for model mesh

For Storage Spec details, please refer to the design doc: https://docs.google.com/document/d/1rYNh93XNMRp8YaR56m-_zK3bqpZhhGKgpVbLPPsi-Ow/edit#

Additional storages/parameters support will come in phase 2.


Tomcli:
@njhill @pvaneck Quick question, is the `schemaPath` only used in Pytorch framework? Right now I have `schemaPath` as optional parameter as part of the storage Spec, but let me know if these kinds of parameters are more framework specific so I can also change it.

In Kserve, the pytorch server assume both model and graph .mar files are under the `model-store` folder instead of pointing to a single file. 
https://console.cloud.google.com/storage/browser/kfserving-examples/models/torchserve/image_classifier;tab=objects?prefix=&forceOnObjectsSortingFiltering=false&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))

In Kserve triton server, it assumes the path is pointing to a folder with structure similar to tensorflow's savedmodel format.
https://console.cloud.google.com/storage/browser/kfserving-examples/models/torchscript?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false

I think the `schemaPath` is useful for backward capability since pytorch didn't have a model saving standard, but I feel like this might be too much to add it as part of the storage spec.

apis/serving/v1beta1/inferenceservice_types.go
```diff
@@ -97,6 +97,25 @@ type PredictorExtensionSpec struct {
 	StorageURI *string `json:"storageUri,omitempty"`
 	// +optional
 	RuntimeVersion *string `json:"runtimeVersion,omitempty"`
+	// Storage Spec for model location
+	// +optional
+	Storage *StorageSpec `json:"storage,omitempty"`
+}
+
+type StorageSpec struct {
```
yuzisun, line 105: Can we discuss this? we do not want to maintain two versions of InferenceService

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +69,55 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+	storageSpec := frameworkSpec.Storage
+	if storageUri == nil {
+		if storageSpec != nil && storageSpec.Path != nil {
+			modelPath = *storageSpec.Path
+			if storageSpec.SchemaPath != nil {
+				schemaPath = *storageSpec.SchemaPath
+			}
+		} else {
+			return "", "", "", "", fmt.Errorf("the InferenceService %v must have either the storageUri or the storage.path", nname)
+		}
```
njhill, line 89: I still think it would be cleaner to invert this one, then you can eliminate an `else` block. Any reason for not doing that?

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +69,55 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
 	return p, nil
 }
 
+// Return secretKey, bucket, modelPath, schemaPath, and error
+func processInferenceServiceStorage(inferenceService *v1beta1.InferenceService, nname types.NamespacedName) (string, string, string, string, error) {
+	var bucket string
+	var modelPath string
+	var schemaPath string
+	secretKey := inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
```
njhill, line 77:
```suggestion
func processInferenceServiceStorage(inferenceService *v1beta1.InferenceService, nname types.NamespacedName) (secretKey, bucket, modelPath, schemaPath string, err error) {
	secretKey = inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
```
This also means that you can remove the parameters from the final `return`.

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +69,55 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+		if storageSpec != nil && storageSpec.Path != nil {
+			return "", "", "", "", fmt.Errorf("the InferenceService %v cannot have both the storageUri and the storage.path", nname)
+		}
+		if !strings.HasPrefix(*storageUri, "s3://") {
+			return "", "", "", "", nil
+		}
+		s3Uri := strings.TrimPrefix(*storageUri, "s3://")
+		urlParts := strings.Split(s3Uri, "/")
+		bucket = urlParts[0]
+		modelPath = strings.Join(urlParts[1:], "/")
```
njhill, line 100: WDYT about using regex here?

njhill (commented):
Thanks @Tomcli, sorry for the delay in re-reviewing!

Tomcli:
thanks @njhill for your reviews

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+		if storageSpec != nil && storageSpec.Path != nil {
+			return "", "", "", "", fmt.Errorf("the InferenceService %v cannot have both the storageUri and the storage.path", nname)
+		}
+		if !strings.HasPrefix(*storageUri, "s3://") {
+			return "", "", "", "", nil
+		}
+		u, err := url.Parse(*storageUri)
+		if err != nil {
+			return "", "", "", "", err
+		}
```
njhill, line 98: @Tomcli kind of a nit but may as well simplify like this:
```suggestion
		u, err := url.Parse(*storageUri)
		if err != nil || u.Scheme != "s3" {
			return "", "", "", "", err
		}
```

config/default/kustomization.yaml
```diff
@@ -94,3 +94,4 @@ resources:
   - ../crd
   - ../rbac
   - ../manager
+namespace: modelmesh-serving
```
njhill, line 97: This looks unrelated to the PR?

njhill:
@Tomcli are you sure we want to include the new `serving.kserve.io_inferenceservices.yaml` file at this point (it's huge)?

May be better to omit it until the other kserve changes are ready? The PR could still be merged, I don't think the new logic will break anything when used with the existing CRD... the storage spec will just appear empty.

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+	storageUri := frameworkSpec.StorageURI
+	storageSpec := frameworkSpec.Storage
+	if storageUri == nil {
+		if storageSpec == nil || storageSpec.Path == nil {
+			return "", "", "", "", fmt.Errorf("the InferenceService %v must have either the storageUri or the storage.path", nname)
+		}
+		modelPath = *storageSpec.Path
+		if storageSpec.SchemaPath != nil {
+			schemaPath = *storageSpec.SchemaPath
+		}
```
njhill, line 87: I think this belongs further down in the `if storageSpec != nil {` section?

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
 	return p, nil
 }
 
+// Return secretKey, bucket, modelPath, schemaPath, and error
+func processInferenceServiceStorage(inferenceService *v1beta1.InferenceService, nname types.NamespacedName) (
+	secretKey, bucket, modelPath, schemaPath string, err error) {
+	secretKey = inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
```
njhill, line 76: Better to put this line near the bottom of the func I think (near line 115):
```go
	if secretKey == "" {
		secretKey = inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
	}
```

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+		if !strings.HasPrefix(*storageUri, "s3://") {
+			return "", "", "", "", nil
+		}
+		u, err := url.Parse(*storageUri)
+		if err != nil {
+			return "", "", "", "", err
+		}
+		bucket = u.Host
+		modelPath = u.Path
+		schemaPath = inferenceService.ObjectMeta.Annotations[v1beta1.SchemaPathAnnotation]
```
njhill, line 101: Move this to end of function? i.e.
```go
if schemaPath == "" {
    schemaPath = inferenceService.ObjectMeta.Annotations[v1beta1.SchemaPathAnnotation]
}

File: pkg/predictor_source/inferenceservice_registry.go [hunk 10]
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+		if storageSpec.Parameters != nil {
+			for k, v := range *storageSpec.Parameters {
+				if k == "bucket" {
+					bucket = v
+				}
+			}
+		}
+	}
+	if secretKey == "" {
+		secretKey = "default"
```
njhill, line 116: I think it makes most sense right now to just leave this as empty if it's not specified

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -68,6 +70,54 @@ func BuildBasePredictorFromInferenceService(isvc *v1beta1.InferenceService) (*v1
...
+	}
+	return secretKey, bucket, modelPath, schemaPath, nil
```
njhill, line 118: With named return params can simplify this:
```suggestion
	return
```

njhill (commented):
Hey @Tomcli sorry I noticed a few more things

Tomcli:
thanks @njhill for your reviews. I updated based on your comments

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -94,24 +141,22 @@ func (isvcr InferenceServiceRegistry) Get(ctx context.Context, nname types.Names
...
-		secretKey := inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
-		if schemaPath, ok := inferenceService.ObjectMeta.Annotations[v1beta1.SchemaPathAnnotation]; ok {
-			p.Spec.SchemaPath = &schemaPath
-		}
-		p.Spec.Storage.S3 = &v1alpha1.S3StorageSource{
-			SecretKey: secretKey,
-			Bucket:    &bucket,
-		}
-		p.Spec.Path = modelPath
+	secretKey, bucket, modelPath, schemaPath, err := processInferenceServiceStorage(inferenceService, nname)
```
njhill, line 145: Thanks @Tomcli looks great now... one more small thing I noticed - might be best for `schemaPath` and `bucket` to be pointers since they are in both CRDs and it would make sense to transfer `nil` values from IS to the corresponding Predictor fields (rather than converting `nil` to a pointer to empty string as is the case now).

Tomcli:
Thanks @njhill I addressed the pointer changes.

njhill (approved):
Thanks @Tomcli

PR #57, opened 2021-10-19
fix: Fix code errors reported by golangci-lint

pugangxa:
#### Motivation
In #53 reported some issues, checked and fixed the ones that are code errors

#### Modifications
Fix bug

#### Result
The code error is fixed, but
- The field alignment is not fixed, not sure whether it's needed and it may have potential impact, maybe just change the golangci-lint rule to ignore it?
- Does not upgrade golangci-lint to 1.42.1 and golang to 1.7 since as mentioned it will be looked separately.

### DIFF
--- a/controllers/modelmesh/etcd.go
+++ b/controllers/modelmesh/etcd.go
@@ -39,8 +39,8 @@ func (m *Deployment) configureMMDeploymentForEtcdSecret(deployment *appsv1.Deplo
 			}
 
 			volumeMountExists := false
-			for i, volumeMount := range container.VolumeMounts {
-				if volumeMount.Name == etcdVolume {
+			for i := range container.VolumeMounts {
+				if container.VolumeMounts[i].Name == etcdVolume {
 					volumeMountExists = true
 					container.VolumeMounts[i].ReadOnly = true
 					container.VolumeMounts[i].MountPath = etcdMountPath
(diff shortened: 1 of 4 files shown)


pugangxa:
Let me know if we want to address them all.

controllers/modelmesh/etcd.go
```diff
@@ -39,11 +39,11 @@ func (m *Deployment) configureMMDeploymentForEtcdSecret(deployment *appsv1.Deplo
...
 
 			volumeMountExists := false
-			for _, volumeMount := range container.VolumeMounts {
+			for i, volumeMount := range container.VolumeMounts {
 				if volumeMount.Name == etcdVolume {
 					volumeMountExists = true
-					volumeMount.ReadOnly = true
-					volumeMount.MountPath = etcdMountPath
+					container.VolumeMounts[i].ReadOnly = true
+					container.VolumeMounts[i].MountPath = etcdMountPath
```
njhill, line 46: May as well avoid the struct copy altogether and also change to:
```go
			for i, _ := range container.VolumeMounts {
				if container.VolumeMounts[i].Name == etcdVolume {
```

njhill (commented):
Thanks @pugangxa, and sorry I missed this earlier. These are legit bugs.

@pvaneck I agree with disabling the `fieldalignment` checks.

controllers/modelmesh/etcd.go
pugangxa: OK, addressed.

njhill (approved):
Thanks @pugangxa!

For future reference, it's best not to modify existing commits that have already been reviewed (apart from rebasing them), just push new commits.

PR #72, opened 2021-11-15
Sync internal changes take 2

njhill:
This is a replay of https://github.com/kserve/modelmesh-serving/pull/69 which was inadvertently squashed rather than merged as individual commits.

**Reminder:** Don't `\lgtm` this PR, we'll `Rebase and merge` when ready


chinhuang007 (approved):
Code and FVT in github actions look good!

PR #73, opened 2021-11-16
feat: Multi-namespace grpc resolver

njhill:
#### Motivation

This is one of the changes that will be needed to make the controller multi-namespace capable

#### Modifications

Update kube grpc client resolver to understand "service.namespace" DNS format and keep track of a resolver per namespaced service.

#### Result

The Kube grpc resolver is namespace aware.

### DIFF
--- a/pkg/mmesh/grpc_resolver_test.go
+++ b/pkg/mmesh/grpc_resolver_test.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/mmesh/grpc_resolver_test.go
```diff
@@ -142,14 +141,6 @@ func reconcile(t *testing.T, kr *KubeResolver) {
 	assert.Nil(t, err)
 }
 
-func makeKubeResolver(namespace string, client client.Client) *KubeResolver {
-	return &KubeResolver{
```
chinhuang007, line 146: Looks makeKubeResolver is still needed in line 78, kr := makeKubeResolver("namespace", mClient)?

chinhuang007: I saw it is moved... so ignore previous comment

chinhuang007 (approved):
/lgtm

PR #75, opened 2021-11-23
refac: Refactor service controller

njhill:
Motivation

The way that the kube `Service` is currently reconciled from the `ConfigMap` uses some hacks to avoid updating the Service when unrelated config changes are made, in particular the use of "special" reconcile `Request` values to indicate whether it was triggered by a config change or not.

In addition:
- The way that the `MMClient` was refreshed is not threadsafe
- Similar hacks are used for reconciling the `ServiceMonitor`
- An "old" Service would not be deleted if the service name was changed in the configuration

Modifications

- Move the general dynamic config handling into its own config package
- Move logic to generate/update a target `ServiceSpec` when the `ConfigMap` changes into this config package and protect with a mutex
- Have the service reconciler compare against this target spec every time
- Make the ServiceMonitor reconcilation more robust and have `ServiceMonitor`s be owner by the corresponding `Service` CRs
- Add more logging

Result

Less hacky and more robust controller logic, ready to be generalized for controlling across multiple namespaces.

(cherry picked from internal commit fab72b11e5d635feabd850c15920a44109a6078e)

### DIFF
--- a/controllers/service_controller.go
+++ b/controllers/service_controller.go
@@ -198,7 +198,7 @@ func (r *ServiceReconciler) reconcileService(ctx context.Context, mms *mmesh.MMS
 
 	namespace := r.ControllerDeployment.Namespace
 	sl := &corev1.ServiceList{}
-	if err := r.List(ctx, sl, client.HasLabels{"wmlserving-service"}, client.InNamespace(namespace)); err != nil {
+	if err := r.List(ctx, sl, client.HasLabels{"modelmesh-service"}, client.InNamespace(namespace)); err != nil {
 		return nil, err, false
 	}
 	var s *corev1.Service
--- a/controllers/service_controller.go
+++ b/controllers/service_controller.go
@@ -33,10 +33,14 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"reflect"
+	"sync"
+
+	"github.com/kserve/modelmesh-serving/pkg/config"
 
 	"github.com/kserve/modelmesh-serving/controllers/modelmesh"
 
-	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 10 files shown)


controllers/service_controller.go
```diff
@@ -169,153 +190,181 @@ func (r *ServiceReconciler) tlsConfigFromSecret(ctx context.Context, secretName
...
-	if exists && (s.Spec.ClusterIP == "None") != headless {
-		r.Log.Info("Deleting/recreating Service because headlessness changed",
-			"service", serviceName, "headless", headless)
-		// Have to recreate since ClusterIP field is immutable
-		if err = r.Delete(ctx, s); err != nil {
-			return err, false
+	namespace := r.ControllerDeployment.Namespace
+	sl := &corev1.ServiceList{}
+	if err := r.List(ctx, sl, client.HasLabels{"wmlserving-service"}, client.InNamespace(namespace)); err != nil {
+		return nil, err, false
```
chinhuang007, line 202: Label probably should be "modelmesh-service"

pkg/mmesh/modelmesh_service.go
```diff
@@ -17,129 +17,240 @@ import (
...
-		changed = true
+	if metricsPort != mms.metricsPort {
+		mms.metricsPort = metricsPort
+		specChange = true
 	}
-	return changed
+
+	if specChange {
+		spec := &v1.ServiceSpec{
+			Selector: map[string]string{"wmlserving-service": mms.name},
```
chinhuang007, line 133: Maybe "modelmesh-service"?

njhill:
Thanks @chinhuang007! Pushed another commit to fix this. Will squash it before merging if tests now look good.

chinhuang007 (approved):
Looks good!

PR #76, opened 2021-11-25
feat: Multi-namespace modelmesh event stream

njhill:
For multi-namespace support various different parts of the controller need to be adapted. This PR updates the ModelMeshEventStream component which streams events from modelmesh's etcd.

It's assumed that the same etcd is used for all namespaces, and that the key prefix is adjusted for non-controller namespaces to be `<root_prefix>/mm_ns/<namespace>/` (as opposed to just `<root_prefix>/` for the controller namespace).

These changes should not affect the single-namespace behaviour or existing tests.


chinhuang007 (approved):
Looks good!

PR #82, opened 2021-12-03
feat: Namespace-aware Service controller

njhill:
The created Service is owned by either the controller `Deployment` or the `Namespace` resource, depending on whether RBAC permissions are present for reading `Namespace`s.


chinhuang007 (approved):
/lgtm

PR #83, opened 2021-12-03
feat: add and use new StorageSpec fields in Predictors

tjohnson31415:
#### Motivation

Adopt the new Storage Spec detailed in [this Design Doc](https://docs.google.com/document/d/1rYNh93XNMRp8YaR56m-_zK3bqpZhhGKgpVbLPPsi-Ow/edit#heading=h.gb4md09ddljj) in Predictors.

This is a follow on to [feat: storage phase 1 for inference service reconciler
](https://github.com/kserve/modelmesh-serving/pull/56)

#### Modifications

- created `apis/serving/common` to have the structs shared between Predictors and InferenceServices
  - currently contains the types for `StorageSpec` and `PredictorStatus` that are now used in both
  - this avoid a circular dependency between the `v1alpha1` and `v1beta1` api packages
- add `StorageSpec` to the Predictor type, deprecating the fields it is replacing
  - part of this includes making `spec.path` optional now that `spec.storage.path` is preferred
- use of `StorageSpec` in Predictors is set up to be mutually exclusive with the previous fields, the model will fail to load and a validation error noted if there is a conflict
- when converting an InferenceService to a Predictor, parse `storageUri` into parameters and have these take precedence over `storage.parameters` before writing them to `storage.parameters` in the Predictor
  - the URI parsing will need to be updated as support for new storage types is added
- update the Predictor Controller to pass arbitrary parameters to the Adapter via the `StorageParameters` field in the `ModelKeyInfo` (instead of just passing `Bucket`)
  
#### Comments

An alternative to the `common` package would be to move the `PredictorStatus ` types into `inferenceservice_types.go` and just have `v1beta1` depend on `v1alpha`.


controllers/predictor_controller.go
```diff
@@ -213,32 +217,62 @@ func (pr *PredictorReconciler) ReconcilePredictor(ctx context.Context, nname typ
...
+		return "Only one of spec.path and spec.storage.path can be specified"
+	}
+
+	if storage.SchemaPath != nil && predictor.Spec.SchemaPath != nil {
+		return "Only one of spec.schemaPath and spec.storage.schemaPath can be specified"
+	}
+
+	// PersistentVolumeClaim is deprecated and was never supported
+	if storage.PersistentVolumeClaim != nil {
+		return "spec.storage.PersistentVolumeClaim is not supported"
```
Tomcli, line 239: Quick question: Are we not going to support PVC for modelmesh? Just wondering why this spec was in the predictor previously. We may still want this feature when adding the new pullman library for single model serving.

Tomcli:
thanks @tjohnson31415 for updating the predictor spec. It looks good to me. Wondering is the predictor going to have explicit spec for http, gcs, azure storage similar to S3 and PVC in the predictor `spec.storage` right now?

controllers/predictor_controller.go
tjohnson31415: My understanding is that `storage.PersistentVolumeClaim` was added to the Predictor spec in the early development of ModelMesh serving, part of the initial draft of the Predictor type. The plan has always been to add support for loading models from PVCs, but S3 support was higher priority and we just never got to it.

Let me know if I can make the code clearer, but the purpose of my change here is to deprecate `storage.PersistentVolumeClaim` in favor of the new fields from the `StorageSpec` that do not have a key referencing the storage type (`storage.path`, `storage.parameters`, `storage.key`, etc.). PVCs are planned to be supported using the new storage spec and implementation with pullman.

tjohnson31415:
My thought was that we should move away from the "old" way that uses the storage-type-specific specs like `storage.S3` and `storage.PersistentVolumeClaim` in favor of the new fields. So support for new storage types would only be via the generalized `StorageSpec` parameters and we wouldn't add any fields to the Predictor spec for each type. `S3` will remain supported for backwards compatibility in the short term, but is deprecated as part of these changes.

Let me know whether or not you agree with that direction!

Tomcli:
Thanks @tjohnson31415 for the explanation. I wasn't aware the history of the predictor storage spec. This makes more sense to me now.
/lgtm

@njhill You can approve this PR if you don't have other comment.

njhill:
Just want to say that I concur with @tjohnson31415's comments above. For PVC, though I was thinking we will support it via the "new" mechanism/syntax, my expectation is that you would not use a secret entry in practice since the config would comprise only one parameter (the PVC name).

So you would have just:
```yaml
    storage:
      path: /path/to/model
      parameters:
        type: pvc
        name: my-pvc
```
or equivalently you could still use
```yaml
    storageUri: pvc://my-pvc/path/to/model
```
(latter would effectively be converted to former but end result would be the same)

Apart from that there are some other things we need to figure out w.r.t. PVC support in the modelmesh case, since we can't connect the storage dynamically to already-running Deployments/Pods.

I'm reviewing this PR's changes more closely now.

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -84,31 +86,50 @@ func processInferenceServiceStorage(inferenceService *v1beta1.InferenceService,
...
 	if secretKey == "" {
 		secretKey = inferenceService.ObjectMeta.Annotations[v1beta1.SecretKeyAnnotation]
 	}
+	if secretKey == "" {
+		if t, ok := parameters["type"]; ok {
+			// default key name for the storage type, used for backwards
+			// compatibility with the changes to the storage spec
+			secretKey = fmt.Sprintf("_%s_default", t)
+		}
+	}
```
njhill, line 132: Trying to remember my prior thought process around this... I don't _think_ we want to set it here, rather just leave it empty and let the defaulting logic happen later as part of the general Predictor processing.

One reason for that is that we want to treat the default secret keys slightly differently (as previously discussed) where it's not an error if they don't exist.

We need to produce an error if neither `secretKey` nor type is provided (where type may be implicit from e.g. URI or deprecated `S3` syntax), but I think that can also be common in the later Predictor processing logic.

Do you see any problem with the above approach?

njhill (commented):
Thanks @tjohnson31415 this looks great overall, please see inline comment/question...

pkg/predictor_source/inferenceservice_registry.go
njhill: (I'm also unsure about the naming of these default entries, but if we are removing that from here it doesn't need to be resolved for this PR!)

tjohnson31415: > just leave it empty and let the defaulting logic happen later as part of the general Predictor processing.
Ah, I thought that the default value for StorageKey would only be injected for an InferenceService, not for a Predictor.

> We need to produce an error if neither secretKey nor type is provided
This would have to come from the Puller, right? To support the custom runtime with no storage helper case, we would not want an error thrown for this in the Controller.

> if we are removing that from here it doesn't need to be resolved for this PR!
So, for this PR, you do not want any default key to be injected, just pass through an empty key? The "later" Predictor processing logic you reference would be in the Puller, right? Is that where you want the default key logic to live?


Agreed about the name of `_%s_default` not being great. That is open for suggestions, maybe something like `_%s_serviceaccount` to tie it back to the SAs, but I don't have a strong suggestion at this point.

njhill: > This would have to come from the Puller, right? To support the custom runtime with no storage helper case, we would not want an error thrown for this in the Controller.

Good point, yes.

> So, for this PR, you do not want any default key to be injected, just pass through an empty key?

Yes

> The "later" Predictor processing logic you reference would be in the Puller, right? Is that where you want the default key logic to live?

Yes and yes, unless you can think of a major reason for not doing that.

njhill (approved):
Thanks @tjohnson31415 lgtm but looks like you have to sign with the DCO per the instructions here: https://github.com/kserve/modelmesh-serving/pull/83/checks?check_run_id=4474073274

Issue #33, opened 2021-09-21
Change ModelMesh Serving controller to work with multiple namespaces


chinhuang007:
The current ModelMesh Serving is namespace scoped, meaning all of its components must exist within a single namespace and only one instance of ModelMesh Serving can be installed per namespace. The idea is to make ModelMesh Serving cluster scoped so that one set of controller components can serve multiple namespaces.

The limitation of having a controller and a set of Serving Runtimes in each namespace means sharing is not possible between namespaces and would cause unnecessarily high resource consumption when ModelMesh Serving is installed in different namespaces.

At a high level, we would like to change Serving Runtimes to be cluster scoped and make the controller to manage resources across namespaces. Here is the [doc ](https://docs.google.com/document/d/12Fscl9yEjDrOAqNEyRhA23MlqUYiHSDOP6XNi5lQYGA/edit?usp=sharing) with more details, including a couple of possible solutions. The doc will be updated whenever decisions are made and implementations are identified.

chinhuang007:
@yuzisun Please help review and suggest/identify a solution that would work naturally in the KServe community and Kube ecosystem. Thanks!

PR #84, opened 2021-12-03
feat: Add support for multiple namespaces

chinhuang007:
Add multi-namespace support so one controller can manage resources
such as ServingRuntimes and Predictors in multiple namespaces

#### Motivation
This PR is to add the multiple namespace support as described in #33, and documented in https://docs.google.com/document/d/12Fscl9yEjDrOAqNEyRhA23MlqUYiHSDOP6XNi5lQYGA/edit#heading=h.6oqyeffyxm7q

#### Modifications
In phase one, we will make the controller to work with multiple namespaces while leave ServingRuntimes as namespaced. The changes are
- controllers to keep track of multiple namespaces
- resolver to work with multiple namespaces
- modelmesh to take namespace as part of service endpoints
- rbac to be cluster scope
- lifecycle support for service, etcd secret, and internal configmap
- install scripts and docs

#### Result
User can install modelmesh-serving once in a controller namespace and deploy ServingRuntimes and Predictors into multiple user namespaces


config/rbac/role.yaml
```diff
@@ -13,14 +13,16 @@
 # limitations under the License.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
njhill, line 16: It would be good to move all the RBAC changes into a separate PR. This will help to make sure everything still works fine with the current permissions and single namespace.

njhill (commented):
Thanks @chauhang, looks good, made a few comments inline.

docs/install/README.md
```diff
@@ -55,11 +57,14 @@ Please be aware that:
 
 For more details see the [built-in runtime configuration](../configuration/built-in-runtimes.md)
 
-In addition, the following resources will be created in the same namespace:
+In addition, the following resources will be created in the control plane namespace:
 
 - `model-serving-defaults` - ConfigMap holding default values tied to a release, should not be modified. Configuration can be overriden by creating a user ConfigMap, see [configuration](../configuration)
 - `tc-config` - ConfigMap used for some internal coordination
 - `storage-config` - Secret holding config for each of the storage backends from which models can be loaded - see [the example](../predictors/)
+- `model-serving-etcd` - Secret providing access to the Etcd cluster - see [instructions](../install/install-script.md#setup-the-etcd-connection-information)
```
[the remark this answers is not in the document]
chinhuang007: I thought "model-serving-etcd" is similar to "storage-config", up to the user to provide. The line above has "`storage-config` - Secret holding config for each of the storage backends from which models can be loaded - see [the example](../predictors/)". So the word "created" seems to include both by code and user.

controllers/servingruntime_controller.go
```diff
@@ -189,7 +256,9 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return RequeueResult, fmt.Errorf("could not determine replicas: %w", err)
 	}
 	mmDeployment.Replicas = replicas
-	if err = mmDeployment.Apply(ctx); err != nil {
+	err = mmDeployment.Apply(ctx)
+
+	if err != nil {
```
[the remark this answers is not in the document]
chinhuang007: Changed back.

controllers/servingruntime_controller.go
```diff
@@ -336,10 +405,11 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
 	for i := range runtimes.Items {
 		rt := &runtimes.Items[i]
 		if runtimeSupportsPredictor(rt, p) {
-			srnns = append(srnns, types.NamespacedName{
+			srnn := types.NamespacedName{
```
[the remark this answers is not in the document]
chinhuang007: Changed back.

pkg/mmesh/modelmesh_events.go
```diff
@@ -123,7 +123,7 @@ func (mes *ModelMeshEventStream) UpdateWatchedService(ctx context.Context,
 				mes.logger.Error(err, "Could not close existing etcd client")
 			}
 		}
-		if err := mes.connectToEtcd(ctx, etcdSecretName); err != nil {
+		if err := mes.connectToEtcd(ctx, etcdSecretName, namespace); err != nil {
```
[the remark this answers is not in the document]
chinhuang007: Removed

pkg/mmesh/modelmesh_events.go
```diff
@@ -202,6 +202,10 @@ func (mes *ModelMeshEventStream) refreshWatches(nw *namespaceWatch, namespace, s
 					if hashIdx > ownerIdx && key[hashIdx] == '-' {
 						// Infer predictor/vmodel and source ids from concrete model id by removing hash suffix
 						sourceId, predictorName := key[ownerIdx:hashIdx], key[:ownerIdx-2]
+						i := strings.LastIndex(namespace, "_")
+						if i > 0 {
+							namespace = namespace[i+1:]
+						}
```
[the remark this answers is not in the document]
chinhuang007: It went into a loop of adding sourceId + "_" without this.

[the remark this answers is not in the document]
njhill: Are you sure that's still the case? There was an oversight in the first version of my PR that caused this, but it was fixed before we merged it. Could you try taking out these changes to verify it's still a problem?
If so then there's another bug somewhere we should fix instead of making these changes.

docs/install/README.md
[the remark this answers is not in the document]
njhill: They are a bit different:
- The etcd secret must be created by the user as a prereq to installation and should be basically static and never (or very rarely) change. However _if_ the `--quickstart` install option is used it will run etcd and create the secret for you.
- The storage config secret does not need to be created by the user, an empty one will be created during installation if it doesn't exist. The user can update this (add/update/delete entries within it) and the system will pick up those changes dynamically. You just need to make sure you add a storage entry to the cert before it can be used by an InferenceService/Predictor. If the `--quickstart` option is used it will also set up minio for you and add an entry for it to the secret.

scripts/install.sh
```diff
@@ -258,4 +264,22 @@ wait_for_pods_ready "-l control-plane=modelmesh-controller"
...
+  kustomize build runtimes --load-restrictor LoadRestrictionsNone > runtimes.yaml
+  cp dependencies/user_ns.yaml .
+  sed -i "s/controller_namespace/${namespace}/g" user_ns.yaml
+
+  for user_ns in "${user_ns_array[@]}"; do
+    if ! kubectl get namespaces $user_ns >/dev/null; then
+      echo "Kube namespace does not exist: $user_ns. Will skip."
+    else 
+      kubectl label namespace ${user_ns} modelmesh-enabled="true"
+      kubectl apply -f user_ns.yaml -n ${user_ns}
```
njhill, line 277: I think we only want to do do this in the `--quickstart` case. Also maybe we should rename it to something like `minio-storage-secret.yaml`?

chinhuang007: Sure. Updated.

docs/install/README.md
```diff
@@ -12,9 +12,11 @@
 
 We provide an install script to quickly run ModelMesh Serving with a provisioned etcd server. This may be useful for experimentation or development but should not be used in production.
 
-## Namespace Scope
+## Cluster Scope
 
-ModelMesh Serving is namespace scoped, meaning all of its components must exist within a single namespace and only one instance of ModelMesh Serving can be installed per namespace. Multiple ModelMesh Serving instances can be installed in separate namespaces within the cluster.
+ModelMesh Serving is cluster scoped, meaning its components can exist in multiple user namespaces which are controlled by one instance of ModelMesh Serving Controller in the control plane namespace. Only one ModelMesh Serving Controller instance can be installed within a Kubernetes cluster.
+
+A namespace label `modelmesh-enabled` needs to be "true" to enable a user namespace for ModelMesh Serving.
```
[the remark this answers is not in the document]
chinhuang007: Made first round of simple updates, assuming the cluster scope mode, to be consistent with Kserve. Will continue to refine.

config/rbac/role.yaml
chinhuang007: Removed RBAC changes.

docs/install/README.md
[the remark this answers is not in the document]
chinhuang007: Thanks for the clarification. If the session is about the resources created by code, not the user, the scenario for multi-namespace seems different. We actually create `model-serving-etcd` by copying one from the controller plane namespace and expect the user to create `storage-config`.

controllers/servingruntime_controller.go
```diff
@@ -113,12 +115,77 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return RequeueResult, fmt.Errorf("could not reconcile the modelmesh type-constraints configmap: %w", err)
 	}
 
-	//TODO(probably) here ... if not controller namespace then
-	//  read etcd secret from controller namespace, replace rootprefix with ns-specific one
-	//  and the create/update etcd secret (with same name) in _this_ namespace
-	//  and include an "ownership" label similar to the tc-config configmap
+	// Reconcile the model mesh cluster config map and etcd secret
```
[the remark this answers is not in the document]
chinhuang007: Good point. Done.

pkg/mmesh/etcd_secret.go
```diff
@@ -0,0 +1,135 @@
...
+	etcdEndpoints := strings.Split(es.EtcdConfig.Endpoints, ",")
+
+	newEtcdEndpoints := ""
+	// For each endpoint, insert controller namespace if not there
+	// TODO: revisit and come up with ideal solution
+	for i := range etcdEndpoints {
+		if !strings.Contains(etcdEndpoints[i], ".") {
+			// This looks like http://etcd:2379, so insert controller namespace before :port
+			parts := strings.Split(etcdEndpoints[i], ":")
+			etcdEndpoints[i] = parts[0] + ":" + parts[1] + "." + es.ControllerNamespace + ":" + parts[2]
```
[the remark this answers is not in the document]
njhill: Something like:
```go
re := regexp.MustCompile("(?:https?://)?([^\\s.:]+)(:\\d+)?")
if se := re.FindStringSubmatchIndex(etcdEndpoints[i]); se != nil {
    etcdEndpoints[i] = fmt.Sprintf("%s.%s%s", etcdEndpoints[i][:se[3]], es.ControllerNamespace, etcdEndpoints[i][se[3]:])
}
```

pkg/mmesh/modelmesh_events.go
```diff
@@ -176,7 +176,7 @@ func (mes *ModelMeshEventStream) refreshWatches(nw *namespaceWatch, namespace, s
 			}
 			if owner, err := ownerIDFromVModelRecord(value); err == nil {
 				encodedNamespace := namespace
-				if owner != "" {
+				if owner != "" && (len(namespace) <= len(owner) || owner+"_" != namespace[:len(owner)+1]) {
```
njhill, line 179: @chinhuang007 this can also be reverted I think... should be no changes at all needed to this file.

main.go
```diff
@@ -190,7 +190,6 @@ func main() {
 	mgrOpts := ctrl.Options{
 		Scheme:                 scheme,
 		MetricsBindAddress:     metricsAddr,
-		Namespace:              ControllerNamespace,
```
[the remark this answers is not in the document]
chinhuang007: Env variable is added in the latest commit. Please take a look.

main.go
```diff
@@ -187,10 +189,40 @@ func main() {
...
+	err = cl.Get(context.Background(), client.ObjectKey{Name: "foo"}, &corev1.Namespace{})
+	clusterScopeEnabled = err == nil || errors.IsNotFound(err)
+
+	if clusterScopeEnabled {
+		if namespaceScope != trueString {
+			setupLog.Info("Controller operating in cluster scope mode, will attempt to watch all namespaces")
+		} else {
+			// Config mismatch, namespace mode with cluster permissions, exit
+			setupLog.Info("In namespace scope mode but controller has cluster scope permissions, exit")
+			os.Exit(1)
```
njhill, line 209: I'm not sure we need to fail in this case, just have more permission than needed.

main.go
```diff
@@ -187,10 +189,40 @@ func main() {
...
+		}
+	} else {
+		if namespaceScope != trueString {
+			// Config mismatch, cluster mode without cluster permissions, exit
+			setupLog.Info("In cluster scope mode but controller has namespace scope permissions, exit")
```
njhill, line 214: `Info` -> `Error`

pkg/mmesh/etcdrangewatcher.go
```diff
@@ -30,17 +30,17 @@ import (
...
-	Password              string `json:"password"`
-	RootPrefix            string `json:"root_prefix"`
-	Certificate           string `json:"certificate"`
-	CertificateFile       string `json:"certificate_file"`
-	ClientKey             string `json:"client_key"`
-	ClientKeyFile         string `json:"client_key_file"`
-	ClientCertificate     string `json:"client_certificate"`
-	ClientCertificateFile string `json:"client_certificate_file"`
-	OverrideAuthority     string `json:"override_authority"`
+	Endpoints             string `json:"endpoints,omitempty"`
```
njhill, line 33: Probably best not to have `omitempty` on this one since it's required (all the others are optional).

pkg/mmesh/etcd_secret.go
```diff
@@ -0,0 +1,91 @@
...
+		if i != 0 {
+			newEtcdEndpoints += ","
+		}
+		newEtcdEndpoints += etcdEndpoints[i]
+	}
+
+	es.EtcdConfig.Endpoints = newEtcdEndpoints
+	es.EtcdConfig.RootPrefix = fmt.Sprintf("%s/mm_ns/%s", es.EtcdConfig.RootPrefix, es.Namespace)
+
+	b, _ := json.Marshal(es.EtcdConfig)
```
njhill, line 86: Sorry forgot to add this in last review - should check for the error here and propagate it back

njhill (commented):
@chinhuang007 just couple more very small things!

main.go
chinhuang007: This might cause unexpected behaviors. The user would expect the controller to watch and reconcile in a single namespace but the controller is acting across all namespaces. I think we should exit and make sure the user/installer syncs up RBAC with install option. What do you think?

chinhuang007: Additionally clusterScopeEnabled will affect how service and SR controllers work internally. For ex, clusterScopeEnabled = true and namespaceScope = true would have different code path from clusterScopeEnabled = false and namespaceScope = true

njhill: @chinhuang007 OK sorry I think the thing I meant to say but forgot to add is that we should also change the later logic to depend on the namespaceScope setting instead of this RBAC check. Basically change the two places where `clusterScopeEnabled` is referenced lower down in this func. Are you ok with the above change if we also do that?

chinhuang007:
@njhill I hope all your comments are addressed. Please take a look.

controllers/servingruntime_controller.go
```diff
@@ -113,12 +115,60 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// replace rootprefix with ns-specific one, and then create/update etcd secret (with same name)
+		// in _this_ namespace and include labels similar to the tc-config configmap
+		s := &corev1.Secret{}
+		err = r.Client.Get(ctx, types.NamespacedName{
+			Name:      etcdSecretName,
+			Namespace: r.ControllerNamespace,
+		}, s)
+		if err != nil {
+			return RequeueResult, fmt.Errorf("Could not get the controller etcd secret: %w", err)
+		}
```
njhill, line 149: nit:
```suggestion
		if err = r.Client.Get(ctx, types.NamespacedName{
			Name:      etcdSecretName,
			Namespace: r.ControllerNamespace,
		}, s); err != nil {
			return RequeueResult, fmt.Errorf("Could not get the controller etcd secret: %w", err)
		}
```

docs/install/README.md
```diff
@@ -55,11 +60,12 @@ Please be aware that:
 
 For more details see the [built-in runtime configuration](../configuration/built-in-runtimes.md)
 
-In addition, the following resources will be created in the same namespace:
+The following resources will be created in the namespaces:
 
 - `model-serving-defaults` - ConfigMap holding default values tied to a release, should not be modified. Configuration can be overriden by creating a user ConfigMap, see [configuration](../configuration)
 - `tc-config` - ConfigMap used for some internal coordination
 - `storage-config` - Secret holding config for each of the storage backends from which models can be loaded - see [the example](../predictors/)
+- `model-serving-etcd` - Secret providing access to the Etcd cluster. It is created by user in the controller namespace - see [instructions](../install/install-script.md#setup-the-etcd-connection-information), and will be automatically created in user namespaces in the cluster scope mode.
```
njhill, line 68: small suggested wording tweak:
```suggestion
- `model-serving-etcd` - Secret providing access to the Etcd cluster. It is created by user in the controller namespace - see [instructions](../install/install-script.md#setup-the-etcd-connection-information), and will be automatically created in user namespaces when in cluster scope mode.
```

pkg/mmesh/etcd_secret.go
```diff
@@ -0,0 +1,99 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	EtcdConnectionKey = "etcd_connection"
+)
```
njhill, line 33: Can use existing constant `modelmesh.EtcdConnectionKey` for this

pkg/mmesh/etcd_secret.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+	}
+	err := es.addData(s)
+	if err != nil {
+		return err
+	}
+	err = cl.Create(ctx, s)
+	if err != nil && errors.IsAlreadyExists(err) {
+		err = cl.Update(ctx, s)
+	}
```
njhill, line 64: The typical state will be already created, so this will cause some unnecessary churn... would be best to do a Get first above before adding the data (which will come from local cache), and then create only if it doesn't exist. If it _does_ exist then just update (or even better check first to see if the data already matches in which case there's nothing to do)

njhill (commented):
Hey @chinhuang007 looks great thanks, just couple of remaining comments (sorry!), hopefully should be pretty quick.

chinhuang007:
@njhill thanks for the review. hopefully addressed your comments.

njhill (approved):
Thanks @chinhuang007 for all your work on this and your patience!

pvaneck (approved):
Thanks, @chinhuang007 and @njhill!
/lgtm

Issue #62, opened 2021-10-27
Increase default gRPC max message size for mlserver runtime


pvaneck:
The default of 4MiB is often too small for many use cases. This should be increased in the mlserver serving runtime config.
This was recently made configurable via https://github.com/SeldonIO/MLServer/pull/317. Should probably be 16Mib to match the modelmesh default.

PR #85, opened 2021-12-04
chore: Update MLServer SR to increase max grpc msg size

pvaneck:
MLServer recently allowed the gRPC max message length to be configurable. This commit uses that variable to increase the max message length to 16 MiB which aligns with ModelMesh's default.

Reference: https://github.com/SeldonIO/MLServer/blob/release/0.5.x/mlserver/settings.py#L39

Closes: #62


njhill (approved):
Thanks @pvaneck!

PR #86, opened 2021-12-06
Sync internal changes 

joerunde:
Note: Do not `/lgtm` this PR!

These are changes to generate TLS secrets for FVT runs.
The key/cert pair was generated with the extra config
```
[ req ]
distinguished_name = req
[ san ]
subjectAltName = DNS:modelmesh-serving,DNS:modelmesh-serving.*,DNS:localhost,IP:0.0.0.0
```


njhill (approved):
Thanks @joerunde

PR #88, opened 2021-12-13
Change RBAC to enable cluster scope

chinhuang007:
Change RBAC to enable cluster scope as the default mode
- role
- role binding
- install default

#### Motivation
Make cluster scope as the default mode for the modelmesh serving controller, to be consistent with kserve

#### Modifications
RBAC roles and role binding
Install default is changed to cluster scope

#### Result
Controller will have cluster role and default installation is in cluster scope mode

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   name: controller-role
 rules:
@@ -21,6 +21,7 @@ rules:
       - ""
     resources:
       - configmaps
+      - namespaces
     verbs:
       - create
       - delete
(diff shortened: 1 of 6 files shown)


config/rbac/role.yaml
```diff
@@ -13,14 +13,15 @@
...
-kind: Role
+kind: ClusterRole
 metadata:
   name: controller-role
 rules:
   - apiGroups:
       - ""
     resources:
       - configmaps
+      - namespaces
```
pvaneck, line 24: I think you may have forgotten the kubebuilder annotation (`+kubebuilder:rbac`) for the `namespaces` resource. Perhaps add it as part of the servingruntime_controller.go rbac annotations?

chinhuang007: Yes, updated role.yaml and put annotations in service_controller.go.

pvaneck:
Thanks @chinhuang007 
/lgtm

njhill (approved):
Thanks @chinhuang007, LGTM

We may want to include in the config alternative RBAC config for the namespace-scope case (perhaps in separate subdir), but can always do that as a follow-on

Issue #80, opened 2021-12-03
Add ServingRuntime field to denote modelmesh compatibility


pvaneck:
With KServe now using ServingRuntimes, we need to make sure that the ModelMesh-Serving runtime selection logic only selects runtimes that are ModelMesh compatible. While currently, KServe leverages the existence of the field `GrpcMultiModelManagementEndpoint` to determine this, let's try to make this more explicit with a dedicated boolean field in the SR spec.

Something like `modelMeshCompatible`, `isMMS`, `supportsModelMesh`? Open to suggestions.

We will also have to make sure the SR controller only creates deployments for accessible runtimes with this field se to true.

PR #89, opened 2021-12-15
feat: Introduce "multiModel" field to SR spec

pvaneck:
#### Motivation
With KServe now using ServingRuntimes, we need to make sure that the ModelMesh-Serving runtime selection logic only selects runtimes that are ModelMesh compatible.

#### Modifications
This adds a field `multi-model` to the ServingRuntime spec which denotes if a given SR is compatible with ModelMesh or not. The default is false. ServingRuntime controller logic was adjusted to account for this field.  An e2e test was added to test this field, and example SRs in docs were updated to include this field.

#### Result

ModelMesh-Serving will only consider runtimes with multiModel = true, and SMS and MMS ServingRuntimes can coexist peacefully.

Close: https://github.com/kserve/modelmesh-serving/issues/80

### DIFF
--- a/apis/serving/v1alpha1/servingruntime_types.go
+++ b/apis/serving/v1alpha1/servingruntime_types.go
@@ -79,6 +79,11 @@ type ServingRuntimePodSpec struct {
 type ServingRuntimeSpec struct {
 	// Model formats and version supported by this runtime
 	SupportedModelTypes []ModelType `json:"supportedModelTypes,omitempty"`
+
+	// Whether this ServingRuntime is intended for multi-model usage or not.
+	// +optional
+	MultiModel *bool `json:"multiModel,omitempty"`
+
 	// Set to true to disable use of this runtime
 	// +optional
 	Disabled *bool `json:"disabled,omitempty"`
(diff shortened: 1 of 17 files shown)


pvaneck:
FYI @njhill If you can take a look.

pvaneck:
@yuzisun What do you think of this `multiModel` field in the SR spec?

apis/serving/v1alpha1/servingruntime_types.go
```diff
@@ -183,3 +188,7 @@ func init() {
 func (sr ServingRuntime) Disabled() bool {
 	return sr.Spec.Disabled != nil && *sr.Spec.Disabled
 }
+
+func (sr ServingRuntime) IsModelMeshCompatible() bool {
```
njhill, line 192: Maybe rename this to `IsMultiModelRuntime`?

njhill (commented):
This looks great thanks @pvaneck, just one naming suggestion. Agree would be good to get @yuzisun's approval too.

apis/serving/v1alpha1/servingruntime_types.go
pvaneck: Yea, I think that name fits better.

yuzisun:
> @yuzisun What do you think of this `multiModel` field in the SR spec?

sgtm, shall we update KServe as well? Also I have updated the KServe go mod you should be able to depend on kserve and share the type now.

pvaneck:
> sgtm, shall we update KServe as well? Also I have updated the KServe go mod you should be able to depend on kserve and share the type now.

Yea, KServe should be updated as well. I will do a PR.

njhill (approved):
Thanks @pvaneck

PR #92, opened 2021-12-17
Enable basic FVTs for multi namespaces

chinhuang007:
Enable basic FVTs for multi namespaces
- add an env variable for controller namespace
- update model-serving-config in the controller namespace
- update FVT workflow to run complete FVTs in user namespace
  in addition to controller namespace

#### Motivation
Enable and run FVTs in both controller and user namespaces in
the default cluster scope mode

#### Modifications
1. Updated fvt code to take an env variable for controller namespace
and configure configmap in the controller namespace.
2. Add/update script in workflow and setup and run FVTs in user namespace 

#### Result
The FVTs should run and pass in multiple namespaces

### DIFF
--- a/fvt/predictor_test.go
+++ b/fvt/predictor_test.go
@@ -119,6 +119,7 @@ var _ = Describe("Predictor", func() {
 					"value": "0",
 				},
 			},
+			"podsPerRuntime": 1,
 		}
 		fvtClient.ApplyUserConfigMap(config)
 
@@ -788,6 +789,7 @@ var _ = Describe("Invalid Predictors", func() {
 					"value": "0",
 				},
 			},
+			"podsPerRuntime": 1,
 		}
 		fvtClient.ApplyUserConfigMap(config)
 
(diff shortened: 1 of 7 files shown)


chinhuang007:
@njhill The same set of FVTs will run in both controller and user namespaces now. Please take a look.

animeshsingh (commented):
Are we waiting kn @njhill or @tjohnson31415 to review for this ?

fvt/predictor_test.go
```diff
@@ -119,6 +119,7 @@ var _ = Describe("Predictor", func() {
 					"value": "0",
 				},
 			},
+			"podsPerRuntime": 1,
```
pvaneck, line 122: I'm assuming these added to avoid hitting the GH Actions environment resource constraints? Are the SR pods in the controller namespace still up  when  the user-namespace tests are run?

chinhuang007: Yes. The SR pods are still up because the predictor test config has scaleToZero=false in order to prevent pods flapping as
			// Predictors are created and deleted

scripts/deploy/iks/test-fvt.sh
```diff
@@ -40,13 +40,24 @@ run_fvt() {
   echo " =====   run standard fvt   ====="
   kubectl get all -n "$SERVING_NS"
   export KUBECONFIG=~/.kube/config
-    
+
+  rm -rf /usr/local/go
+  wget https://go.dev/dl/go1.16.4.linux-amd64.tar.gz
```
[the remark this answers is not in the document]
chinhuang007: Thanks for pointing out the version discrepancy. Just synced with the Github Actions FVT workflow and completed a toolchain pipeline run.

PR #94, opened 2021-12-25
config: upgrade the cert manager resource version

haoxins:
Same as https://github.com/kserve/kserve/commit/df512fdf943303fb21d9fd8d3aaff818c5705851


pvaneck (commented):
Thanks!
/lgtm

PR #95, opened 2021-12-25
Config: Upgrade the K8s APIs to v1

haoxins:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/


pvaneck (commented):
Thanks!
/lgtm

PR #98, opened 2022-01-05
refactor: be explicit in StorageParams having string values only

tjohnson31415:
#### Motivation

Small follow-up to https://github.com/kserve/modelmesh-serving/pull/83. Initially I was thinking that we'd pass a JSON struct through under `StorageParams`, but, after consideration, decided that we should just pass through the parameters from the Predictor's `storage.parameters` field as-is without any transformation.

[These changes](https://github.com/kserve/modelmesh-runtime-adapter/pull/12) in the puller parse a `map[string]string`  from the storage parameters, so the two sides of the communication match up for runtimes using the puller.

#### Modifications

- refactor `StorageParams` to be a `map[string]string` instead of `map[string]interface{}`
- remove unneeded `tranformParameters` (sic 😅 ) function

### DIFF
--- a/controllers/predictor_controller.go
+++ b/controllers/predictor_controller.go
@@ -255,10 +255,10 @@ func validatePredictor(predictor *api.Predictor) string {
 
 // passed in ModelInfo.Key field of registration requests
 type ModelKeyInfo struct {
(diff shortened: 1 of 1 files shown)


njhill (approved):
LGTM, thanks!

Issue #99, opened 2022-01-05
Import ServingRuntime and InferenceService types from KServe


pvaneck:
Types for these resources are currently defined in two places: the `kserve` repository and the `modelmesh-serving` repository. Let's investigate importing these types from KServe and using those within modelmesh-serving as a step towards unification.

yuzisun:
I have updated the KServe `go.mod` you should be able to import the types now.

pvaneck:
Closed via #140 and https://github.com/kserve/modelmesh-serving/pull/146

Issue #90, opened 2021-12-16
Sync ServingRuntime spec between MM and KServe


pvaneck:
Currently the CRDs for ServingRuntimes are generated in two places: the `kserve/kserve` repo and the `kserve/modelmesh-serving` repo. Although it'd be preferred to have a single source, for now, let's ensure that the specs are synchronized. This way a user won't run into issues when using ServingRuntimes in the different contexts.

List of changes;
- KServe has [readinessProbes](https://github.com/kserve/kserve/blob/1754f0870864f315a032d415c0de82a71d60d81c/pkg/apis/serving/v1alpha1/servingruntime_types.go#L50) as a part of the `Container` struct.
  - [x] Needs PR in `ModelMesh-Serving`
- MM has PR for introducing `multiModel` field to SR (https://github.com/kserve/modelmesh-serving/pull/89)
  - [x] Once merged, needs PR in `KServe' 
- KServe has PR for introducing `autoPlace` field to ModelType/Framework (https://github.com/kserve/kserve/pull/1948)
  - [x] Once merged, needs PR in `ModelMesh-Serving'  
- KServe ServingRuntime and InferenceService Predictor Spec [Renames](https://github.com/kserve/kserve/pull/1956)
  - Rename Framework/ModelType to ModelFormat on InferenceService CR
  - Rename Framework/ModelType to SupportedModelFormat on ServingRuntime CR

PR #100, opened 2022-01-05
chore: Sync KServe SR CRD with MM SR CRD

pvaneck:
#### Motivation

The ServingRuntime CRDs between KServe and ModelMesh-Serving have a few differences. This commit makes the necessary changes so that specs reflect the same things. Eventually, we'll aim to remove this duplication via https://github.com/kserve/modelmesh-serving/issues/99.

Part of: https://github.com/kserve/modelmesh-serving/issues/90 

#### Modifications

- In ServingRuntime, `SupportedModelTypes`  was renamed to `SupportedModelFormats`, and references to `SupportedModelTypes` were also changed throughout the code and documentation.
- `ModelType` struct was added to predictor and used as the type for modelType.
- `SupportedModelFormat` struct added to SR types which is similar to `ModelType` but with an additional `AutoSelect` field that is used in determining if SR is eligible for automatic placement for a given model format.

#### Result

Same spec produced for ServingRuntime CRDs between both KServe and ModelMesh.

**Note:** This maintains the use of `modelType` in the predictor spec so that no existing `Predictor` CR breaks. However, this is a breaking change in the context of ServingRuntimes as `SupportedModelTypes` is removed from the spec in favor of `SupportedModelFormats`. Presumably this should be okay as the spec if still listed as alpha, but let me know if you think backwards compatibility should be maintained.


njhill (commented):
Thanks @pvaneck this looks good.

Shouldn't there also be changes to the mm controller to honor the `autoSelect` value?

I also wonder whether we could have a new global config parameter that controls the default value of this (with its own default being false)? I guess an issue with doing that is the fact that this configmap is only used by the mm controller but `autoSelect` also affects logic in the main kserve controller?

Re backwards compatibility, I'd guess that this is ok given that it's alpha like you say and also I doubt many folks are using this yet, would be good to get @yuzisun's view/agreement though.

pvaneck:
Thanks, @njhill 
> Shouldn't there also be changes to the mm controller to honor the autoSelect value?

Ah, yea, there should be. Can have it done in a followup pr. Right now MM is treating all model formats having autoSelect = true.

> I also wonder whether we could have a new global config parameter that controls the default value of this (with its own default being false)? I guess an issue with doing that is the fact that this configmap is only used by the mm controller but autoSelect also affects logic in the main kserve controller?

I think that would probably be nice to have, but yea that config map issue arises. I suppose ModelMesh would benefit more from this config in order to maintain current behavior, so we could potentially make it MM specific.  Otherwise, we might need to add it to both the KServe and MM configmaps (with associated logic), then consolidate once we start merging controllers. 

Though on a side note, I think soon we'll need to have the ModelMesh controller check for and access the KServe configmap in order to get the `defaultDeploymentMode` value for allowing ModelMesh deployments for ISvcs without the deploymentMode annotation.

Issue #96, opened 2021-12-28
Sync InferenceService CR to ModelMesh Serving controller


yuzisun:
**Describe the bug**

```
{"level":"error","ts":1640666407.3959372,"logger":"controller.predictor","msg":"Reconciler error","reconciler group":"serving.kserve.io","reconciler kind":"Predictor","name":"example-sklearn-isvc","namespace":"isvc_modelmesh-serving","error":"failed to fetch CR from kubebuilder cache for predictor example-sklearn-isvc: No valid InferenceService predictor framework found","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/root/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.2/pkg/internal/controller/controller.go:227"}
```

**To Reproduce**

1. Create InferenceService CR with modelmesh annotation
2. ModelMesh controller fails to reconcile with above error

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- Version: latest modelmesh-controller from master branch

**Additional context**

Issue #97, opened 2021-12-29
Model fails to load when using InferenceService storageUri


yuzisun:
**Describe the bug**

**To Reproduce**

1. Create the following example InferenceService CR
```
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: mnist 
  annotations:
    serving.kserve.io/deploymentMode: ModelMesh
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    sklearn:
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```
2. Model fails to load on model puller sidecar

```
2021-12-29T12:56:32.918Z	DEBUG	MLServer Adapter.MLServer Adapter Server	Listing objects from s3	{"bucket": "modelmesh-example-models", "prefix": "/sklearn/mnist-svm.joblib"}
2021-12-29T12:56:32.960Z	DEBUG	MLServer Adapter.MLServer Adapter Server	Ignore downloading s3 object matching part of the path	{"bucket": "modelmesh-example-models", "prefix": "/sklearn/mnist-svm.joblib", "s3_path": "sklearn/mnist-svm.joblib"}
2021-12-29T12:56:32.960Z	ERROR	MLServer Adapter.MLServer Adapter Server.Load Model	Failed to pull model from storage	{"model_id": "mnist__isvc-04bd33724e", "error": "rpc error: code = Unknown desc = Failed to pull model from storage due to error: no objects found for path '/sklearn/mnist-svm.joblib'"}
github.com/kserve/modelmesh-runtime-adapter/internal/proto/mmesh._ModelRuntime_LoadModel_Handler
	/opt/app/internal/proto/mmesh/model-runtime_grpc.pb.go:175
google.golang.org/grpc.(*Server).processUnaryRPC
	/root/go/pkg/mod/google.golang.org/grpc@v1.36.0/server.go:1217
google.golang.org/grpc.(*Server).handleStream
	/root/go/pkg/mod/google.golang.org/grpc@v1.36.0/server.go:1540
google.golang.org/grpc.(*Server).serveStreams.func1.2
	/root/go/pkg/mod/google.golang.org/grpc@v1.36.0/server.go:878
```

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

**Additional context**

Phelan164:
@yuzisun I tried to deploy InferenceService and got error
[...]

PR #101, opened 2022-01-07
feat: Update InferenceService reconciliation logic

pvaneck:
The ModelMesh controller can now reconcile InferenceServices using the new Model Spec in the predictor. Also, an issue was fixed with there being a leading slash in the model path when an InferenceService StorageURI was parsed.  This was causing  issues in the adapter, preventing models from being loaded as seen in #97.

Closes: #96, #97
Related: #90 

#### Result

Users can now successfully deploy an InferenceService using the predictor Model Spec such as the following:

```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn 
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```


pkg/predictor_source/inferenceservice_registry_test.go
```diff
@@ -20,9 +20,127 @@ import (
...
+						Version: &formatVersion,
+					},
+					Runtime: &runtimeName,
+				},
+			},
+		},
+	}
+	predictor, err := BuildBasePredictorFromInferenceService(inferenceService)
+	assert.NoError(t, err)
+	assert.Equal(t, predictor.Spec.Model.Type.Name, formatName)
```
tjohnson31415, line 76:
```suggestion
	assert.Equal(t, formatName, predictor.Spec.Model.Type.Name)
```

pkg/predictor_source/inferenceservice_registry_test.go
```diff
@@ -20,9 +20,127 @@ import (
...
+			Predictor: v1beta1.InferenceServicePredictorSpec{
+				SKLearn: &v1beta1.PredictorExtensionSpec{
+					StorageURI: &uri,
+				},
+			},
+		},
+	}
+	predictor, err := BuildBasePredictorFromInferenceService(inferenceService)
+	assert.NoError(t, err)
+	assert.Equal(t, predictor.Spec.Model.Type.Name, "sklearn")
```
tjohnson31415, line 105:
```suggestion
	assert.Equal(t, "sklearn", predictor.Spec.Model.Type.Name)
```

pkg/predictor_source/inferenceservice_registry_test.go
```diff
@@ -20,9 +20,127 @@ import (
...
+					ModelFormat: v1beta1.ModelFormat{
+						Name: formatName,
+					},
+				},
+			},
+		},
+	}
+	predictor, err := BuildBasePredictorFromInferenceService(inferenceService)
+	assert.NoError(t, err)
+	assert.Equal(t, predictor.Spec.Model.Type.Name, formatName)
```
tjohnson31415, line 47:
```suggestion
	assert.Equal(t, formatName, predictor.Spec.Model.Type.Name)
```

tjohnson31415 (commented):
Looks pretty good to me.

There is a small thing about the tests with the order of expected and actual in `assert.Equal` ([REF](https://pkg.go.dev/github.com/stretchr/testify/assert#Equal)).

For the deprecation of the model-type specific fields, we may want to be stricter in their exclusivity to avoid any potential confusion. In particular, maybe there should be errors if:
- `model` and the `serving.kserve.io/servingRuntime` annotation both exist
- `model` and a type-specific extension block both exist in `InferenceServicePredictorSpec`

pvaneck:
Thanks @tjohnson31415 for the review. Good calls on the assert argument orders and the stricter field usage checks. I have updated the PR per you suggestions.

yuzisun:
@pvaneck I have updated the [kserve go mod dependencies](https://github.com/kserve/kserve/blob/master/go.mod) now you should be able to depend on kserve with`go get github.com/kserve/kserve@172933a487b5a7ee521c79f9ce6440476d0b4dcf` and remove the `InferenceServiceSpec` type in ModelMesh repo, for `InferenceServiceStatus` I am thinking to introduce a new struct `ModelStatus` so we can share the full type.

Phelan164:
@pvaneck I now trying modelmesh for model orchestration purpose. I tried your inference service spec above and got below issue.
`error: error validating "test.yaml": error validating data: ValidationError(InferenceService.spec.predictor): unknown field "model" in io.kserve.serving.v1beta1.InferenceService.spec.predictor; if you choose to ignore these errors, turn validation off with --validate=false`
My env:  kserve v0.7.0 (latest release one)
I have 1 question: is this compatible with v0.7.0 or need latest kserve deploy from source code?

pvaneck:
> @pvaneck I have updated the [kserve go mod dependencies](https://github.com/kserve/kserve/blob/master/go.mod) now you should be able to depend on kserve with`go get github.com/kserve/kserve@172933a487b5a7ee521c79f9ce6440476d0b4dcf` and remove the `InferenceServiceSpec` type in ModelMesh repo, for `InferenceServiceStatus` I am thinking to introduce a new struct `ModelStatus` so we can share the full type.

Thanks, Dan. I can start looking into that, but want to merge this in for now to get ISvc logic working.

> @pvaneck I now trying modelmesh for model orchestration purpose. I tried your inference service spec above and got below issue. `error: error validating "test.yaml": error validating data: ValidationError(InferenceService.spec.predictor): unknown field "model" in io.kserve.serving.v1beta1.InferenceService.spec.predictor; if you choose to ignore these errors, turn validation off with --validate=false` My env: kserve v0.7.0 (latest release one) I have 1 question: is this compatible with v0.7.0 or need latest kserve deploy from source code?

Hey, @Phelan164. The above InferenceService spec is pretty new and has not been released just yet, so you would need to install from the master branch of KServe (it will be released soon though!) With KServe v0.7.0, you can still use the spec with the framework keys. E.g.:

```
kubectl apply -f - <<EOF
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    sklearn:
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
EOF
```
However, as mentioned in this PR, there are some issues with ISvcs with the latest version of modelmesh, so need to make sure the v0.7.0 version is applied:

```shell
kubectl apply -f https://github.com/kserve/modelmesh-serving/releases/download/v0.7.0/modelmesh-quickstart-dependencies.yaml
for i in 1 2 3 4 5 ; do kubectl apply -f https://github.com/kserve/modelmesh-serving/releases/download/v0.7.0/modelmesh.yaml && break || sleep 15; done
for i in 1 2 3 4 5 ; do kubectl apply -f https://github.com/kserve/modelmesh-serving/releases/download/v0.7.0/modelmesh-runtimes.yaml && break || sleep 15; done

```

njhill (approved):
Thanks @pvaneck

PR #108, opened 2022-01-14
feat: Add autoSelect ServingRuntime logic

pvaneck:
## Motivation
In each supported model format of a ServingRuntime, a user can specify autoSelect as true to indicate that the given ServingRuntime can be used for automatic placement of a model of that format. Kserve uses this setting for automatic model placement decisions, but MM doesn't yet currently

## Modifications
Logic was added to have MM only consider SRs with model formats containing autoSelect as true when finding compatible runtimes.

The out-of-the-box ServingRuntimes now have "autoSelect: true" on all of their supported formats, so there is no behavior change for users.

## Results
MM respects the autoSelect field in each supported model format entry.

### DIFF
--- a/controllers/modelmesh/model_type_labels.go
+++ b/controllers/modelmesh/model_type_labels.go
@@ -47,9 +47,12 @@ func GetServingRuntimeSupportedModelTypeLabelSet(rt *api.ServingRuntime) StringS
 
 	// model type labels
 	for _, t := range rt.Spec.SupportedModelFormats {
-		set.Add("mt:" + t.Name)
-		if t.Version != nil {
-			set.Add("mt:" + t.Name + ":" + *t.Version)
+		// only include model type labels when autoSelect is true
+		if t.AutoSelect != nil && *t.AutoSelect {
+			set.Add("mt:" + t.Name)
+			if t.Version != nil {
+				set.Add("mt:" + t.Name + ":" + *t.Version)
+			}
 		}
 	}
 	// runtime label
--- a/controllers/modelmesh/model_type_labels_test.go
+++ b/controllers/modelmesh/model_type_labels_test.go
@@ -24,6 +24,7 @@ import (
 func TestGetServingRuntimeSupportedModelTypeLabelSet(t *testing.T) {
 	version_semver := "12345.312.2"
 	version_someString := "someString"
+	autoSelectVal := true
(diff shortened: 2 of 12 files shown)


controllers/modelmesh/model_type_labels_test.go
```diff
@@ -53,14 +57,17 @@ func TestGetServingRuntimeSupportedModelTypeLabelSet(t *testing.T) {
 
 	expectedLabels := []string{
 		"mt:type1",
-		"mt:type3",
+		"mt:type2",
```
chinhuang007, line 60: Should this match l52 as type3?

pvaneck: Since `type3` doesn't have the `AutoSelect: true` label, it is not expected to be one of the returned labels. However, `mt:type2` should be included since a type2 model format does have the `AutoSelect: true` label (this wasn't here initially, but probably should have been)

njhill (approved):
Looks great, thanks @pvaneck

controllers/modelmesh/model_type_labels_test.go
chinhuang007: Hmm, I am not sure the logic in GetServingRuntimeSupportedModelTypeLabelSet is correct then.

I might misunderstand this, but I thought the following should create one entry as supported model type, including the version number.
{
         Name:       "type2",
         Version:    &version_semver,
         AutoSelect: &autoSelectVal,
},

The current code in model_type_labels.go would create two entries, meaning the version number is irrelevant. Is this expected?

controllers/modelmesh/model_type_labels.go
```diff
@@ -47,9 +47,12 @@ func GetServingRuntimeSupportedModelTypeLabelSet(rt *api.ServingRuntime) StringS
 
 	// model type labels
 	for _, t := range rt.Spec.SupportedModelFormats {
-		set.Add("mt:" + t.Name)
-		if t.Version != nil {
-			set.Add("mt:" + t.Name + ":" + *t.Version)
+		// only include model type labels when autoSelect is true
+		if t.AutoSelect != nil && *t.AutoSelect {
+			set.Add("mt:" + t.Name)
+			if t.Version != nil {
```
chinhuang007, line 53: Related to my other comment, maybe check t.Version first and add only one entry to set?

PR #109, opened 2022-01-19
Update Keras model example and docs

tedhtchang:
#### Motivation
Update Keras doc and examples to reflect the changes in the updated minio images

### DIFF
--- a/config/example-predictors/example-keras-mnist-predictor.yaml
+++ b/config/example-predictors/example-keras-mnist-predictor.yaml
@@ -1,0 +1,24 @@
+# Copyright 2022 IBM Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


docs/model-types/README.md
```diff
@@ -2,6 +2,7 @@
 
 By leveraging existing third-party model servers, we support a number of standard ML model formats out-of-the box, with more to follow. Currently supported model types:
 
+- [keras] (keras.md)
 - [LightGBM](lightgbm.md)
```
chinhuang007, line 6: Maybe remove the blank character between "]" and "("?

config/example-predictors/example-keras-mnist-predictor.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2021 IBM Corporation
```
pvaneck, line 1: 2021 -> 2022

docs/model-types/keras.md
```diff
@@ -35,7 +35,7 @@ metadata:
 spec:
   modelType:
     name: keras
-  path: tensorflow-model/keras-mnist/mnist.h5
+  path: keras-models/mnist.h5
```
pvaneck, line 38: This doesn't match the path above. Which is the correct one?

docs/model-types/README.md
pvaneck: Right, also, a nit: capitalize the K in keras.

pvaneck (changes requested):

pvaneck (approved):

PR #118, opened 2022-01-28
feat: Allow installation using link to config archive

pvaneck:
A user can now optionally pass in a link to an archive file of the configs using the `-p` option of the install.sh script.

This is more of a nice-to-have that allows some flexibility with installation options.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -32,7 +32,7 @@ function showHelp() {
   echo
   echo "Flags:"
   echo "  -n, --namespace                (required) Kubernetes namespace to deploy ModelMesh Serving to."
-  echo "  -p, --install-config-path      Path to local model serve installation configs. Can be ModelMesh Serving tarfile or directory."
+  echo "  -p, --install-config-path      Path to installation configs. Can be a local ModelMesh Serving config tarfile/directory or a URL to a config tarfile."
   echo "  -d, --delete                   Delete any existing instances of ModelMesh Serving in Kube namespace before running install, including CRDs, RBACs, controller, older CRD with serving.kserve.io api group name, etc."
   echo "  -u, --user-namespaces          Kubernetes namespaces to enable for ModelMesh Serving"
   echo "  --quickstart                   Install and configure required supporting datastores in the same namespace (etcd and MinIO) - for experimentation/development"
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -196,14 +196,23 @@ info "Setting kube context to use namespace: $namespace"
...
+if [[ -n $install_config_path ]]; then
+  if [[ -f $install_config_path ]] && [[ $install_config_path =~ \.t?gz$ ]]; then
+    tar -xf "$install_config_path"
+    cd "$(basename "$(basename $install_config_path .tgz)" .tar.gz)"
+  elif [[ $install_config_path =~ ^http.+ ]]; then
+    if [[ $install_config_path =~ \.t?gz$ ]]; then
+      curl -L $install_config_path -O
+      filename=$(basename $install_config_path)
+      tar -xf "$filename"
+      cd "$(basename "$(basename $filename .tgz)" .tar.gz)"
```
chinhuang007, line 208: Could you provide an example? I tried with `./scripts/install.sh -p https://github.com/kserve/modelmesh-serving/archive/refs/tags/v0.7.0.tar.gz -n modelmesh-serving` and got an error, "cd: v0.7.0: No such file or directory"

chinhuang007: I suppose the script can work with special built tgz and tar.gz files which contain just the config folder contents. It is not clear though based on the flag description, ".... Can be ModelMesh Serving tarfile or directory."

scripts/install.sh
```diff
@@ -196,14 +196,23 @@ info "Setting kube context to use namespace: $namespace"
...
+    else
+      die "Provided URL should be a .tgz or tar.gz archive file"
+    fi
+  elif [[ -d $install_config_path ]]; then
+    cd "$install_config_path"
```
chinhuang007, line 213: When I provide the "ModelMesh Serving directory", it fails later with "cd: default: No such file or directory".

There are two possible solutions:
1. change the description to "ModelMesh Serving config directory"
2. add "cd config" wherever applicable

scripts/install.sh
chinhuang007: Are you suggesting to replace the source tar.gz which is commonly for the entire project? Shall we name it more specifically like config-v0.8.0-rc0.tar.gz?

pvaneck: Ah no, that will still remain with just a different download endpoint, but I can see how it might be confusing when they have the same names. So, I agree, let's name it more specifically.

chinhuang007 (approved):
/lgtm

PR #119, opened 2022-01-28
Fix install issues for user namespaces on Mac

chinhuang007:
Fix install issues for user namespaces on Mac

#### Motivation
Fix mac os install issues

#### Modifications
Use sed i.bak

#### Result
User can install in mac os


pvaneck (commented):
Thanks, @chinhuang007 
/approve

PR #120, opened 2022-01-28
feat: FVT for inference service.

### LINKED ISSUE #106: Add functional tests for InferenceServices
The InferenceService CRD is the primary interface for KServe, so we want to ensure that model deployments on ModelMesh work as expected when users use InferenceService CRD.

I think for this, we just need some basic tests:

1. Deploy an Isvc with the `"serving.kserve.io/deploymentMode": "ModelMesh"` annotation, and ensure that it becomes ready, and you can perform inference.

Would like to test the following formats:

```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    sklearn:
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```
And also using the new format:
```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc2
  annotations:
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn 
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```
And once the new storage spec is in (https://github.com/kserve/kserve/pull/1899):
```yaml
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
spec:
  predictor:
    sklearn:
      storage:
        key: localMinIO
        path: sklearn/mnist-svm.joblib
        # schemaPath: null
        parameters:
          bucket: modelmesh-example-models
```

The `run-fvt.yml` GitHub actions workflow will probably need to be updated to install the InferenceService CRD onto the minikube cluster before ModelMesh-Serving is installed.

ScrapCodes:
closes #106

### DIFF
--- a/fvt/helpers.go
+++ b/fvt/helpers.go
@@ -34,6 +34,14 @@ func NewPredictorForFVT(filename string) *unstructured.Unstructured {
 	return p
 }
 
+func NewIsvcForFVT(filename string) *unstructured.Unstructured {
+	p := DecodeResourceFromFile(isvcSamplesPath + filename)
+	uniqueName := MakeUniquePredictorName(p.GetName())
+	p.SetName(uniqueName)
+
+	return p
+}
+
 // to enable tests to run in parallel even when loading from the same Predictor
 // sample
 func MakeUniquePredictorName(base string) string {
@@ -56,6 +64,18 @@ func CreatePredictorAndWaitAndExpectLoaded(predictorManifest *unstructured.Unstr
 	resultingPredictor := WaitForLastStateInExpectedList("activeModelState", []string{"Pending", "Loading", "Standby", "FailedToLoad", "Loading", "Loaded"}, watcher)
 	ExpectPredictorState(resultingPredictor, true, "Loaded", "", "UpToDate")
 	return resultingPredictor
+}
+
+func CreateIsvcAndWaitAndExpectReady(isvcManifest *unstructured.Unstructured) *unstructured.Unstructured {
+	isvcName := isvcManifest.GetName()
+	By("Creating inference service " + isvcName)
+	watcher := fvtClient.StartWatchingIsvcs(metav1.ListOptions{FieldSelector: "metadata.name=" + isvcName}, defaultTimeout)
+	defer watcher.Stop()
+	fvtClient.CreateIsvcExpectSuccess(isvcManifest)
+	By("Waiting for inference service" + isvcName + " to be 'Ready'")
+	// ISVC does not have the status field set initially.
+	resultingIsvc := WaitForIsvcReady(watcher)
+	return resultingIsvc
 }
 
 func CreatePredictorAndWaitAndExpectFailed(predictorManifest *unstructured.Unstructured) *unstructured.Unstructured {
@@ -166,6 +186,50 @@ func ExpectPredictorFailureInfo(obj *unstructured.Unstructured, reason string, h
 		Expect(err).To(BeNil())
 		Expect(time.Since(actualTime) < time.Minute).To(BeTrue())
 	}
+}
+
+func WaitForIsvcReady(watcher watch.Interface) *unstructured.Unstructured {
+	ch := watcher.ResultChan()
+	isReady := false
+	var obj *unstructured.Unstructured
+	var isvcName string
+
+	timeout := time.After(predictorTimeout)
+	done := false
+	for !done {
+		select {
+		// Exit the loop if InferenceService is not ready before given timeout.
+		case <-timeout:
+			done = true
+		case event, ok := <-ch:
+			if !ok {
+				// the channel was closed (watcher timeout reached)
+				done = true
+				break
+			}
+			obj, ok = event.Object.(*unstructured.Unstructured)
+			Expect(ok).To(BeTrue())
+			isvcName = GetString(obj, "metadata", "name")
+			conditions, exists := GetSlice(obj, "status", "conditions")
+			if !exists {
+				time.Sleep(time.Second)
+				continue
+			}
+			for _, condition := range conditions {
+				conditionMap := condition.(map[string]interface{})
+				if conditionMap["type"] == "Ready" {
+					if conditionMap["status"] == "True" {
+						isReady = true
+						done = true
+						break
+					}
+				}
+			}
+
+		}
+	}
+	Expect(isReady).To(BeTrue(), "Timeout before InferenceService '%s' ready", isvcName)
+	return obj
 }
 
 // Waiting for predictor state to reach the last one in the expected list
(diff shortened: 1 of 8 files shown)


fvt/helpers.go
```diff
@@ -115,15 +139,15 @@ func UpdatePredictorAndWaitAndExpectFailed(predictorManifest *unstructured.Unstr
...
 
 	By("Waiting for predictor " + predictorName + "'s target model state to be 'FailedToLoad'")
 	resultingPredictor := WaitForLastStateInExpectedList("targetModelState", []string{"", "Loading", "FailedToLoad"}, watcher)
-	ExpectPredictorState(resultingPredictor, true, "Loaded", "FailedToLoad", "BlockedByFailedLoad")
+	ExpectState(resultingPredictor, true, "Loaded", "FailedToLoad", "BlockedByFailedLoad")
 	return resultingPredictor
 }
 
-func ExpectPredictorState(obj *unstructured.Unstructured, available bool, activeModelState, targetModelState, transitionStatus string) {
+func ExpectState(obj *unstructured.Unstructured, available bool, activeModelState, targetModelState, transitionStatus string) {
```
pvaneck, line 150: Hmm, not sure if status fields will be reflected in the object (in the case of InferenceServices) when this is being checked. Will need to check.

pvaneck (commented):
Left some comments inline.

Some tips when running FVTs locally:

- If running full tests, modelmesh-serving should be installed using the `--fvt` flag with the install.sh script. Don't think this matters if just running isvc tests, though.
- Should set the namespace being used in an environment variable before running the tests: `export NAMESPACE=modelmesh-serving`
- Run the tests using  `go test -v ./fvt -ginkgo.v -ginkgo.progress -ginkgo.failFast -test.timeout 40m`
- To only run the isvc tests,  add an `F` in front of the respective `Specify` and `It` blocks. So for your case, in `isvc_fvt_test.go`, you'd have:

```
FSpecify("Preparing the cluster for inference service tests", func() {
...
FIt("should successfully load a model", func() {
```

pvaneck:
Also, need to install the inferenceservice CRD in the Github Actions environment before ModelMesh Serving is installed.

Can probably just add a `kubectl apply -f https://raw.githubusercontent.com/kserve/kserve/master/test/crds/serving.kserve.io_inferenceservices.yaml` [here](https://github.com/kserve/modelmesh-serving/blob/main/.github/workflows/run-fvt.yml#L53).

For your local testing, I think just installing ModelMesh Serving then the above crd yaml should be fine. No need for Knative and Istio.

fvt/helpers.go
ScrapCodes: Hi Paul, it is not reflected. I am unsure why this is so. If I try to get the object by `kubectl` then I can see the status field.

pvaneck: Yea, I just tried attempting to reuse `WaitForLastStateInExpectedList` and it was giving me issues because sometimes a status field exists and sometimes it doesn't. Was thinking maybe its easier just to use something like this:

```golang
func WaitForIsvcReady(watcher watch.Interface) *unstructured.Unstructured {
	ch := watcher.ResultChan()
	isReady := false
	var obj *unstructured.Unstructured
	var isvcName string

	timeout := time.After(predictorTimeout)
	done := false
	for !done {
		select {
		// Exit the loop if InferenceService is not ready before given timeout.
		case <-timeout:
			done = true
		case event, ok := <-ch:
			if !ok {
				// the channel was closed (watcher timeout reached)
				done = true
				break
			}
			obj, ok = event.Object.(*unstructured.Unstructured)
			Expect(ok).To(BeTrue())
			isvcName = GetString(obj, "metadata", "name")
			conditions := GetSlice(obj, "status", "conditions")
			for _, condition := range conditions {
				conditionMap := condition.(map[string]interface{})
				if conditionMap["type"] == "Ready" {
					if conditionMap["status"] == "True" {
						isReady = true
						done = true
						break
					}
				}
			}

		}
	}
	Expect(isReady).To(BeTrue(), "Timeout before InferenceService '%s' ready", isvcName)
	return obj
}
```

Note: `GetSlice` is an added function in `utils.go` which is very similar to `GetString`. 

WDYT about just checking isvc status ready condition like the above?

ScrapCodes: The object received at this stage looks like :
```
Obj: &unstructured.Unstructured{Object:map[string]interface {}{"apiVersion":"serving.kserve.io/v1beta1", "kind":"InferenceService", "metadata":map[string]interface {}{"annotations":map[string]interface {}{"serving.kserve.io/deploymentMode":"ModelMesh", "serving.kserve.io/secretKey":"localMinIO"}, "creationTimestamp":"2022-03-02T10:22:18Z", "generation":1, "managedFields":[]interface {}{map[string]interface {}{"apiVersion":"serving.kserve.io/v1beta1", "fieldsType":"FieldsV1", "fieldsV1":map[string]interface {}{"f:metadata":map[string]interface {}{"f:annotations":map[string]interface {}{".":map[string]interface {}{}, "f:serving.kserve.io/deploymentMode":map[string]interface {}{}, "f:serving.kserve.io/secretKey":map[string]interface {}{}}}, "f:spec":map[string]interface {}{".":map[string]interface {}{}, "f:predictor":map[string]interface {}{".":map[string]interface {}{}, "f:model":map[string]interface {}{".":map[string]interface {}{}, "f:modelFormat":map[string]interface {}{".":map[string]interface {}{}, "f:name":map[string]interface {}{}}, "f:storageUri":map[string]interface {}{}}}}}, "manager":"fvt.test", "operation":"Update", "time":"2022-03-02T10:22:18Z"}}, "name":"example-sklearn-isvc2-wz5g9", "namespace":"modelmesh-serving", "resourceVersion":"1549528", "uid":"48d6e243-ee52-40c0-ad04-a5899703b0df"}, "spec":map[string]interface {}{"predictor":map[string]interface {}{"model":map[string]interface {}{"modelFormat":map[string]interface {}{"name":"sklearn"}, "storageUri":"s3://modelmesh-example-models/sklearn/mnist-svm.joblib"}}}}}
```
It does not have field condition.

ScrapCodes: Thank you, this works!

pvaneck:
@ScrapCodes  Looks like the FVTs managed to pass here: https://github.com/kserve/modelmesh-serving/actions/runs/1932155743 but not the latest one. Was there anything different with the code that time versus the latest?

ScrapCodes:
Hi @pvaneck, no difference, looks like test is flaky!

pvaneck (approved):
/lgtm

PR #121, opened 2022-02-01
feat: Enable gcs and http for ISvc storage

pvaneck:
#### Motivation

We want to enable users to deploy InferenceServices using models hosted on GCS or online through an http(s) URI.

#### Modifications

Since the model-serving-puller should now (or soon) be able to support `gcs` and `http` types, we let these through when parsing the storageURI. 

#### Result

Users can now deploy ISVCs using GCS, both unauthenticated and authenticated. 

Example resources I've tested:

Unauthenticated:
```yaml
kind: InferenceService
metadata:
  name: pv-sklearn-mnist
  annotations:
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn 
      storageUri: "gs://pv-kfserving/sklearn/mnist-svm.joblib"
```
Authenticated:
```yaml
kind: InferenceService
metadata:
  name: pv-sklearn-mnist2
  annotations:
    serving.kserve.io/secretKey: pvGCS
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn 
      storageUri: "gs://pv-protected/sklearn/mnist-svm.joblib"
```

Users can now also deploy basic HTTP(S) based models, but this seems to currently only be limited to single model files and doesn't yet support archives.

Example I tested with:

```yaml
kind: InferenceService
metadata:
  name: pv-sklearn-mnist3
  annotations:
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn 
      storageUri: "https://storage.googleapis.com/pv-kfserving/sklearn/mnist-svm.joblib"
```

These were all loaded successfully and inference was performed successfully.


chinhuang007 (approved):
/lgtm

PR #122, opened 2022-02-01
docs: Miscellaneous minor doc fixes

pvaneck:
- Updated quickstart to clone release branch
- Renamed disable to disabled which is the correct field name
- Added note about updating namespace prep script to release doc
- Removed misleading InferenceService storage spec docs as the new storage spec PR hasn't been merged yet. These can be re-added once the spec is actually updated.


chinhuang007 (approved):
/lgtm

PR #123, opened 2022-02-11
chore: Propagate GrpcMaxMessageSize to REST proxy

pvaneck:
In order to keep the max gRPC message sizes aligned between ModelMesh and the REST-proxy, this PR takes the config value used by ModelMesh and passes it in to the REST-proxy as an env variable.


njhill (approved):
:+1:

PR #133, opened 2022-02-27
mod: Upgrade to ginkgo v2

haoxins:


go.mod
```diff
@@ -10,15 +10,16 @@ require (
...
 	github.com/moverest/mnist v0.0.0-20160628192128-ec5d9d203b59
-	github.com/onsi/ginkgo v1.16.4
-	github.com/onsi/gomega v1.15.0
+	github.com/onsi/ginkgo/v2 v2.1.3
+	github.com/onsi/gomega v1.18.1
 	github.com/operator-framework/operator-lib v0.8.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.52.0
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.7.0
 	github.com/tommy351/goldga v0.3.0
```
pvaneck, line 19: This should be updated to v0.5.0 for ginkgo v2 support. I believe that's why the unit tests are failing.

pvaneck (commented):
Thanks for the upgrading ginkgo! I left an inline comment that should get the unit tests passing.

go.mod
haoxins: Thanks, and tests passed.

pvaneck:
Thanks, @haoxins!
/lgtm
/approve

PR #134, opened 2022-03-11
Fix scripts for toolchain FVT

chinhuang007:
Fix scripts for toolchain FVT so the InferenceService CRD
from kserve will be installed and removed
Also increase FVT retry to 2


pvaneck (approved):
/lgtm
/approve

PR #135, opened 2022-03-17
chore: sync small internal changes

anhuong:
#### Motivation

In an effort to rebase our internal wml-serving repo on the kserve/modelmesh-serving repo, we are syncing some of the differences internally externally. All of the below changes already exist internally.

#### Modifications

- Set MLServer GRPC message size to unlimited
   - This is because the MLServer GRPC message size is already limited by the `MM_SVC_GRPC_MAX_MSG_SIZE`. This allows us to only have to worry about one variable that limits the message size.
-  Refactor checking digest runtime images to see if they match the provided digest into a separate method
- Remove clientChange for rest port, do not need to reconnect to mmesh client if rest port changes
- Set changed flag if endpoints change in predictor reconciler
   - In the predictor reconciler, we should update the Predictor Status if/when just the resolved endpoints change. This in theory could happen independently of other things changing since it's tied to the service lifecycle.
   - Make sure to set the changed flag in the updatePredictorStatusFromVModel function if either http or grpc endpoint has change.
   - Copied changes from [internal commit](https://github.ibm.com/ai-foundation/wml-serving/commit/c5806427bb3c02f38a329d7a8e2eced66eb060ea)

#### Result

Cleaner code that aligns with internal version.

Signed-off-by: Anh-Uong <anh.uong@ibm.com>

### DIFF
--- a/config/runtimes/mlserver-0.x.yaml
+++ b/config/runtimes/mlserver-0.x.yaml
@@ -55,7 +55,8 @@ spec:
         # Set server addr to localhost to ensure MLServer only listen inside the pod
         - name: MLSERVER_HOST
           value: "127.0.0.1"
-        # Increase gRPC max message size to 16 MiB to support larger payloads
+        # Increase gRPC max message size to support larger payloads
+        # Unlimited because it will be restricted at the model mesh layer
         - name: MLSERVER_GRPC_MAX_MESSAGE_LENGTH
           value: "-1"
       resources:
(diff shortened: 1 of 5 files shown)


pvaneck:
Thanks, @anhuong. Current FVT failures are due to some known flakiness with recently added Isvc tests. Other tests passed fine. I will push this through and investigate the flakiness separately.

/unhold
/approve

PR #139, opened 2022-03-29
Add performance test to e2e toolchain

chinhuang007:
Add performance test to e2e toolchain

#### Motivation
Make sure the performance is not degraded

#### Modifications
Add a step to the end to end toolchain that runs every night to run performance tests, developed under [modelmesh-performance](https://github.com/kserve/modelmesh-performance) repo.

#### Result
Performance tests will be run nightly and the step will fail if performance is worse than the threshold.


pvaneck (commented):
/approve

pvaneck (commented):
/lgtm

PR #140, opened 2022-03-30
refac: use KServe types for InferenceServices

pvaneck:
#### Motivation

With the InferenceService status now supporting the needed ModelMesh status fields, we can now use the type definitions from KServe instead of defining duplicate types here.

#### Modifications

This commit imports the kserve go package and uses the types defined there for InferenceServices and their related fields.
Old InferenceService type definitions are removed.

#### Result
InferenceService types are now only defined in one place (i.e. the kserve repo)

Ref: https://github.com/kserve/modelmesh-serving/issues/99

Signed-off-by: Paul Van Eck <pvaneck@us.ibm.com>


pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -258,23 +310,62 @@ func (isvcr InferenceServiceRegistry) UpdateStatus(ctx context.Context, predicto
...
+		inferenceService.Status.Components[v1beta1.PredictorComponent] = componentStatus
+
+		inferenceService.Status.SetCondition(v1beta1.PredictorReady, &apis.Condition{
+			Type:   v1beta1.PredictorReady,
+			Status: corev1.ConditionTrue,
+		})
+		inferenceService.Status.SetCondition(v1beta1.IngressReady, &apis.Condition{
+			Type:   v1beta1.IngressReady,
+			Status: corev1.ConditionTrue,
+		})
```
njhill, line 358: What's the thinking behind this? We don't currently configure an ingress in the MM case right?

njhill (commented):
This looks great, thanks @pvaneck.

We also need to make some changes for the change to use `v1.Container`, were you thinking to do that in this PR or a different one? I'm happy to add some changes for that too.

pkg/predictor_source/inferenceservice_registry.go
pvaneck: Hah, this tripped me up for a while as I was initially trying to set the Ready condition to True directly. In KServe, the Ready condition for an ISvc is dependent on a list of conditions mentioned [here](https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1beta1/inference_service_status.go#L242) with `IngressReady` being one of those conditions. So, every time I call `InferenceService.Status.SetCondition`, if the prerequisite conditions are True, then `Ready` is set to True. 

But I can probably find an alternative way to set the PredictorReady and Ready conditions without going through the `SetCondition` function. I agree that the IngressReady condition doesn't make much sense in ModelMesh.

pvaneck:
> We also need to make some changes for the change to use v1.Container, were you thinking to do that in this PR or a different one? I'm happy to add some changes for that too.

Yea, planned on handling ServingRuntime transitioning into a dedicated PR.

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -258,23 +310,62 @@ func (isvcr InferenceServiceRegistry) UpdateStatus(ctx context.Context, predicto
...
 	inferenceService.ObjectMeta = predictor.ObjectMeta
-	inferenceService.Status.PredictorStatus = predictor.Status
+
+	inferenceService.Status.ModelStatus.TransitionStatus = v1beta1.TransitionStatus(predictor.Status.TransitionStatus)
+	inferenceService.Status.ModelStatus.ModelRevisionStates = &v1beta1.ModelRevisionStates{
+		ActiveModelState: v1beta1.ModelState(predictor.Status.ActiveModelState),
+		TargetModelState: v1beta1.ModelState(predictor.Status.TargetModelState),
+	}
+	inferenceService.Status.ModelStatus.ModelCopies = &v1beta1.ModelCopies{
+		FailedCopies: predictor.Status.FailedCopies,
```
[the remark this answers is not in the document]
njhill: Well I was still hoping to support `LoadingCopies` at some point but to implement in a reasonable way requires some nontrivial rework of modelmesh internals, so will stay as a "possible future addition" for now.

pvaneck:
Had to update the FVT GH Actions workflow to free up disk space since we were running into space issues.

njhill (approved):

PR #141, opened 2022-03-30
feat: add support and FVTs for new built-in OVMS runtime

tjohnson31415:
#### Motivation

Adds support for a new built-in runtime using OpenVINO Model Server (OVMS). From [the OVMS docs](https://docs.openvino.ai/latest/ovms_what_is_openvino_model_server.html):

> OpenVINO Model Server (OVMS) is a high-performance system for serving machine learning models. It is based on C++ for high scalability and optimized for Intel solutions, so that you can take advantage of all the power of the Intel® Xeon® processor or Intel’s AI accelerators and expose it over a network interface. OVMS uses the same architecture and API as [TensorFlow Serving](https://github.com/tensorflow/serving), while applying OpenVINO for inference execution. Inference service is provided via gRPC or REST API, making it easy to deploy new algorithms and AI experiments.

The OVMS supports running models in [OpenVINO's Intermediate Representation (IR) format](https://docs.openvino.ai/latest/openvino_docs_MO_DG_IR_and_opsets.html), as well as ONNX models.

#### Modifications

- adds OVMS runtime as a built-in runtime
- updates rpcConfigs so that a TFS Predict request does not need the `mm-vmodel-id` header (it is extracted from the request itself)
- adds basic FVTs for loading and inferring with an OpenVINO IR model (using an MNIST model converted from the existing ONNX model)

Adding an additional runtime pushed up against the resource limits of the FVT tests build which only has 2 CPU and 8Gi allocated to it. Some changes were made to get FVTs to a passing state:
- added printing of node details and events on FVT failure
- deploy runtimes with memory reduced to 512Mi
- do not include testing of user namespace installs in the FVTs (having a second namespace doubles the resource requirements when scale-to-zero is disabled...)


config/runtimes/ovms-1.x.yaml
```diff
@@ -0,0 +1,55 @@
...
+  grpcDataEndpoint: "port:8001"
+
+  containers:
+    - name: ovms
+      image: ovms-1:replace
+      args:
+        - --port=8001
+        - --rest_port=8888
+        - --config_path=/models/model_config_list.json # must match the default value in the ovms adapter server
+        - --file_system_poll_wait_seconds=0
```
njhill, line 41: Could add a comment here that we disable polling because the adapter explicitly calls the reload API.

config/runtimes/ovms-1.x.yaml
```diff
@@ -0,0 +1,55 @@
...
+        - --rest_bind_address=127.0.0.1
+      resources:
+        limits:
+          cpu: "5"
+          memory: 1Gi
+        requests:
+          cpu: 500m
+          memory: 1Gi
+  builtInAdapter:
+    serverType: "ovms"
```
njhill, line 52: nit: no need to quote

config/runtimes/ovms-1.x.yaml
```diff
@@ -0,0 +1,55 @@
...
+    runtimeManagementPort: 8888
+    memBufferBytes: 134217728
```
njhill, line 54: Unrelated to this PR but while we are making breaking updates to this CRD it might be nice to change this to be a string in the same form as is used in `resources`.

config/runtimes/ovms-1.x.yaml
```diff
@@ -0,0 +1,55 @@
...
+  grpcEndpoint: "port:8085"
+  grpcDataEndpoint: "port:8001"
+
+  containers:
+    - name: ovms
+      image: ovms-1:replace
+      args:
+        - --port=8001
+        - --rest_port=8888
+        - --config_path=/models/model_config_list.json # must match the default value in the ovms adapter server
```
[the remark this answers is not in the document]
tjohnson31415: OVMS will retry reading the file a few times and then crash:
```
$ docker run openvino/model_server:latest --config_path non-existent.json
[2022-03-31 06:54:13.375][1][serving][info][server.cpp:105] OpenVINO Model Server 2022.1.2b4bec50
[2022-03-31 06:54:13.375][1][serving][info][server.cpp:106] OpenVINO backend 2022.1.0.643
[E:] [BSL] found 0 ioexpander device
E: [xLinkUsb] [    653390] [ovms] usb_find_device_with_bcd:266	Library has not been initialized when loaded
[2022-03-31 06:54:13.390][1][modelmanager][info][modelmanager.cpp:110] Available devices for Open VINO: CPU, GNA
[2022-03-31 06:54:13.391][1][serving][error][modelmanager.cpp:640] Failed to debug-read fileconfig
[2022-03-31 06:54:13.391][1][modelmanager][error][modelmanager.cpp:674] Configuration file is invalid non-existent.json
[2022-03-31 06:54:14.393][1][serving][error][modelmanager.cpp:640] Failed to debug-read fileconfig
[2022-03-31 06:54:14.394][1][modelmanager][error][modelmanager.cpp:674] Configuration file is invalid non-existent.json
[2022-03-31 06:54:15.395][1][modelmanager][error][modelmanager.cpp:159] Couldn't start model manager
[2022-03-31 06:54:15.395][1][serving][error][server.cpp:186] ovms::ModelManager::Start() Error: Configuration file not found or cannot open
```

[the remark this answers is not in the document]
njhill: OK thanks, just thinking about the startup ordering, i.e. if the adapter hasn't finished starting and created the file yet.

njhill (commented):
Thanks @tjohnson31415! Will wait for the remaining TODOs before final review.

njhill (approved):
Thanks @tjohnson31415!

I guess we'll hold off merging this until the other CRD update/sync PRs are in (hopefully soon)?

njhill:
As discussed we'll merge this now, no need to wait for the other CRD changes.

PR #142, opened 2022-04-08
feat: Add TotalCopies in Predictor Status spec

pvaneck:
**Motivation** 

`TotalCopies` is a recently added field to the InferenceService status which tracks the number
of copies/replicas of the model. We should start using it.

**Modifications**

`TotalCopies` was added to the `PredictorStatus`, and the Predictor CRD was regenerated.
The InferenceService registry now propagates this field.

**Result**

`TotalCopies` is now available in the `Predictor` and `InferenceService` statuses.

### DIFF
--- a/apis/serving/common/common_types.go
+++ b/apis/serving/common/common_types.go
@@ -137,11 +137,11 @@ type PredictorStatus struct {
 	// +optional
 	GrpcEndpoint string `json:"grpcEndpoint"`
 
-	//TODO TBC whether or not these are exposed here
+	// Total number copies of this predictor's models that are currently loaded
+	// +kubebuilder:default=0
+	TotalCopies int `json:"totalCopies"`
 
-	//	// How many copies of this predictor's models are currently loaded - NOT YET SUPPORTED
-	//	// +kubebuilder:default=0
-	//	LoadedCopies int `json:"loadedCopies"`
+	//TODO TBC whether or not this is exposed here
 	//	// How many copies of this predictor's models are currently loading - NOT YET SUPPORTED
 	//	// +kubebuilder:default=0
 	//	LoadingCopies int `json:"loadingCopies"`
--- a/apis/serving/common/common_types.go
+++ b/apis/serving/common/common_types.go
@@ -145,7 +145,7 @@ type PredictorStatus struct {
 	//	// +kubebuilder:default=0
 	//	LoadingCopies int `json:"loadingCopies"`
 
-	// Total number copies of this predictor's models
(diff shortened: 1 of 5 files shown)


pvaneck:
@njhill  Does the modelmesh side need to be updated to support this? Seems like the function that performs the counts never increments the `mmeshapi.ModelStatusInfo_LOADED`  count as all the CopyStatuses appear to be `UNKNOWN`.

controllers/predictor_controller.go
```diff
@@ -537,6 +537,10 @@ func (pr *PredictorReconciler) updatePredictorStatusFromVModel(status *common.Pr
 	//	status.LoadingCopies, status.LoadedCopies, status.FailedCopies = counts[0], counts[1], counts[2]
 	//	changed = true
 	//}
+	if counts[1] != status.TotalCopies {
+		status.TotalCopies = counts[1]
+		changed = true
+	}
```
njhill, line 543: Maybe combine into one if like the commented code above?

njhill:
> Does the modelmesh side need to be updated to support this? Seems like the function that performs the counts never increments the `mmeshapi.ModelStatusInfo_LOADED` count as all the CopyStatuses appear to be `UNKNOWN`.

@pvaneck this is the reason for only adding Total for now and not also Loaded/Loading. So we should make the `counts` slice one longer for a total number and increment it in the `countModelCopyStates` function regardless of the `CopyStatus`.

apis/serving/common/common_types.go
```diff
@@ -138,14 +138,17 @@ type PredictorStatus struct {
...
 	//TODO TBC whether or not these are exposed here
-
 	//	// How many copies of this predictor's models are currently loaded - NOT YET SUPPORTED
 	//	// +kubebuilder:default=0
 	//	LoadedCopies int `json:"loadedCopies"`
 	//	// How many copies of this predictor's models are currently loading - NOT YET SUPPORTED
 	//	// +kubebuilder:default=0
 	//	LoadingCopies int `json:"loadingCopies"`
 
+	// Total number copies of this predictor's models
```
njhill, line 148: nit:
```suggestion
	// Total number of copies of this predictor's models
```

I'm still kind of on the fence about whether to include the failed ones or not. I guess like you suggested we should in any case keep this consistent with what's done for the knative case (i.e. if some number of the scaled pods are failing)

njhill (approved):
Thanks @pvaneck, I guess we can easily adjust this later if we decide differently about including failed copies in total copies.

PR #144, opened 2022-04-14
feat: set labels and annotations with user configmap

anhuong:
#### Motivation

Allows users to set labels and annotations for ServingRuntime pods via the `model-serving-config` ConfigMap. This can be used for setting things like network policy for the runtime pods, which was requested by a user.

#### Modifications

- Sets SR deployment PodSpec labels and annotations from the user ConfigMap
- Removes labels `network-policy` and `component`
- Sets new key delimiter for reading ConfigMaps to `::` to allow for dot notation label values like `app.kubernetes.io`


controllers/modelmesh/runtime.go
```diff
@@ -333,3 +333,32 @@ func (m *Deployment) addPassThroughPodFieldsToDeployment(deployment *appsv1.Depl
 
 	return nil
 }
+
+func (m *Deployment) configureRuntimePodSpecAnnotations(deployment *appsv1.Deployment) error {
+
+	if deployment.Spec.Template.Annotations == nil {
+		deployment.Spec.Template.Annotations = make(map[string]string)
```
njhill, line 340: Small thing but for these you could just have this be
```go
    deployment.Spec.Template.Annotations = m.AnnotationsMap
```
and then have the loop below in an `else` block. Same for the labels.

config/default/config-defaults.yaml
```diff
@@ -51,3 +51,6 @@ storageHelperResources:
 serviceAccountName: ""
 metrics:
   enabled: true
+labels:
+  # Required in Watson Cloud envs to allow outbound access to etcd
+  network-policy: allow-egress
```
njhill, line 56: Probably best to remove this completely, we can add it internally.

pkg/config/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 	Metrics     PrometheusConfig
 	ScaleToZero ScaleToZeroConfig
 
+	Labels      map[string]string
+	Annotations map[string]string
```
njhill, line 81: I think these should have slightly more specific names, maybe
`runtimePodLabels`, `runtimePodAnnotations`. Otherwise it's not clear what these really refer to.

njhill (commented):
Thanks @anhuong this looks good. As well as the inline comments could you add the new parameters to the docs [here](https://github.com/kserve/modelmesh-serving/tree/main/docs/configuration)?

config/internal/base/deployment.yaml.tmpl
```diff
@@ -45,8 +45,6 @@ spec:
         app.kubernetes.io/instance: modelmesh-controller
         app.kubernetes.io/managed-by: modelmesh-controller
         app.kubernetes.io/name: modelmesh-controller
-        # Required in Watson Cloud envs to allow outbound access to etcd
-        network-policy: allow-egress
```
anhuong, line 49: @njhill Want to make sure, would this affect kserve users of model-mesh?

docs/configuration/README.md
```diff
@@ -23,26 +23,28 @@ data:
...
+| `grpcMaxMessageSizeBytes`                  | The max number of bytes for the gRPC request payloads (\*\*\*\* see below)                            | `16777216` (16MiB)                                                    |
+| `restProxy.enabled`                        | Enables the provided REST proxy container being deployed in each ServingRuntime deployment            | `true`                                                                |
+| `restProxy.port`                           | Port on which the REST proxy to serve REST requests                                                   | `8008`                                                                |
+| `runtimePodLabels`                         | `metadata.labels` to be added to all ServingRuntime pods                                              | (\*\*\*\*\*See below)                                                 |
+| `runtimePodAnnotations`                    | `metadata.annotations` to be added to all `ServingRuntime` pods                                       | [Cloudpak billing](<(../production-use/cloudpak-billing.md)>) details |
```
anhuong, line 47: Need to remove link here and check on default values

config/internal/base/deployment.yaml.tmpl
njhill: @anhuong I don't think so

njhill:
@anhuong you'll also remove the hardcoded `component: model-mesh` label right?

njhill (approved):
Thanks @anhuong

PR #145, opened 2022-04-14
deps: Update many dependencies

njhill:
Includes resolution of etcd client vulnerability CVE-2018-1099

Note that we are keeping the Kubernetes libraries at 22.x level since there's now a transitive dependency on knative via the main kserve libraries and that's incompatible with Kube >= 23.0.


njhill:
Unfortunately there's an incompatibility of the latest kubernetes libraries with knative. This repo now depends on knative via the dependency on kserve for the shared CRDs. Knative types are used in the InferenceService status.

Per other references found to this it sounds like there's no straightforward way around it (see https://github.com/Kong/kubernetes-ingress-controller/issues/2072#issuecomment-995334293). Looks like they aren't plannign to update to newer Kubernetes library versions for some time. For now will walk back the versions to the latest compatible version.

njhill:
Moved Kubernetes dependency back to latest 0.22.x (v0.22.8), builds and FVT look good now.

pvaneck (commented):
Thanks!
/lgtm
/approve

PR #147, opened 2022-04-20
deps: Avoid reference to older etcd dependency in go.sum

njhill:
#### Motivation

Dependabot is reporting a vulnerability due to an old version of `go.etcd.io/etcd` referenced in go.sum. This came in via an old version of `k8s.io/apiserver` which in turn came in via `knative.dev/serving` which doesn't yet have a newer version.

#### Modifications

Add replace directive for `k8s.io/apiserver` to go.sum to pin to the same v0.22.8 version as other `k8s.io` dependencies.


tjohnson31415:
As advertised
/lgtm
/approve

PR #148, opened 2022-04-20
chore: mirror ServingRuntime type changes from kserve/kserve

tjohnson31415:
#### Motivation

We are working to align the CRDs between this repo and kserve/kserve. KServe will be the source-of-truth. We want this repo to pull directly from KServe, but, for this PR, the updates to ServingRuntimes are copied manually.

#### Modifications

Bring in CRD changes already merged and proposed changes by copying servingruntime_types.go from https://github.com/kserve/kserve/pull/2147/commits/2b38b73d5887800dbc77bf13ffbd662d537eaf1b.

Serving Runtime type/CRD changes:
- embed corev1.Container instead of limited spec
- ClusterServingRuntimes
- env field for BuiltInAdapter
- add Volumes to SR spec
- remove enum validation of ServerType
- add ProtcolVersions to ServingRuntime specs

Other changes:
- generated code updated with `make generate manifests`
- minor code changes to get code to build and pass tests

PR #149, opened 2022-04-25
feat: support env overrides for the adapter

tjohnson31415:
#### Motivation

We are working to align CRDs between this repo and kserve/kserve. In https://github.com/kserve/modelmesh-serving/pull/148, we mirrored updates to the ServingRuntime CRDs. This PR implements use of BuiltInAdapter.Env.

#### Modifications

- allow values in BuiltInAdapter.Env to override environment variables added by the controller

### DIFF
--- a/controllers/modelmesh/runtime.go
+++ b/controllers/modelmesh/runtime.go
@@ -187,7 +187,9 @@ func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error
 
 		// the puller and adapter containers are the same image and are given the
 		// same resources
-		builtInAdapterContainer.Resources = *m.PullerResources
+		if m.PullerResources != nil {
+			builtInAdapterContainer.Resources = *m.PullerResources
+		}
 
 		builtInAdapterContainer.VolumeMounts = append(volumeMounts, corev1.VolumeMount{
 			Name:      ConfigStorageMount,
@@ -195,7 +197,7 @@ func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error
 			ReadOnly:  true,
 		})
 
-		builtInAdapterContainer.Env = []corev1.EnvVar{
+		defaultEnvVars := []corev1.EnvVar{
 			{
 				Name:  "ADAPTER_PORT",
 				Value: "8085",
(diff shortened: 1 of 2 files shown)


controllers/modelmesh/runtime_test.go
```diff
@@ -73,30 +95,54 @@ func TestOverlayMockRuntime(t *testing.T) {
 				Spec: corev1.PodSpec{
 					Containers: []corev1.Container{
 						{
-							Name: "mm",
+							Name: adapterType,
```
pvaneck, line 98: nit: Not a big deal, but shouldn't this be maintained as "mm" to coincide with the actual initial container name?

controllers/modelmesh/runtime.go
```diff
@@ -229,8 +231,13 @@ func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error
...
 			},
-			{}, {}, // allocate larger array to avoid reallocation
-		}[:7]
+		}
+
+		// To avoid reallocation, count the maximum number of env vars there could be
+		numVarsDefault := len(defaultEnvVars) // count of default env vars
+		numVarsOptional := 2                  // count of envs that may be added from annotations below
+		builtInAdapterContainer.Env = make([]corev1.EnvVar, numVarsDefault, numVarsDefault+numVarsOptional+len(rt.Spec.BuiltInAdapter.Env))
+		copy(builtInAdapterContainer.Env, defaultEnvVars)
```
njhill, line 240: Not sure it's worth the extra complexity at this point, especially since this now does involve a separate array allocation. Maybe just revert the changes above here? Feel free to also remove the existing larger-array alloc logic if you don't like it.

controllers/modelmesh/runtime.go
```diff
@@ -187,15 +187,17 @@ func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error
 
 		// the puller and adapter containers are the same image and are given the
 		// same resources
-		builtInAdapterContainer.Resources = *m.PullerResources
+		if m.PullerResources != nil {
```
njhill, line 190: I don't think this should ever be nil here, might be better to let it fail with deref error since it would imply a bug. Or otherwise at least add an error log in the nil case?

njhill (commented):
Thanks @tjohnson31415 looks good just minor comments

controllers/modelmesh/runtime.go
tjohnson31415: The main reason for the change is to support the test changes. Adding use of `BuiltInAdapter` to the tests triggered a nil pointer exception here where the Deployment is not created with all fields being set. I will add a test helper to create a complete `modelmesh.Deployment` for the unit tests (including the test logger I added).

Also I know the controller doesn't support it, but it is valid for a container to not have any resources specified (BestEffort QoS). If it was supported, it would be useful for small dev/test environments

controllers/modelmesh/runtime_test.go
tjohnson31415: Ah, I think I changed this in the wrong spot... With the addition of BuiltInAdapter, there should be a container with the same name as the `ServerType` (though this is checked well before the real code path would reach this point), so the runtime container is the thing that should have a matching name.

Nice catch.

pvaneck (commented):
/lgtm

njhill (approved):

PR #150, opened 2022-04-25
feat: implement additions to the runtime pod spec

tjohnson31415:
#### Motivation

We are working to align CRDs between this repo and kserve/kserve. In https://github.com/kserve/modelmesh-serving/pull/148, we mirrored updates to the ServingRuntime CRDs. This PR provides implementations related to the changes to the Pod spec included in the runtime: allowing additional Volumes to be specified, using the full `v1.Container` spec.

These additions allow more freedom in how a Runtime Deployment can be configured, but this also opens up more potential for conflicts with the containers injected by the controller. For this reason, this PR also adds some constraints and validation on ServingRuntimes, but attempts to mostly keep the current behavior.

#### Modifications

- small refactors to gather and expose variables with names for internal Volumes
- use full `v1.Container` from the CRD when generating a runtime Deployment
- add validation of the runtime container specs now that they are more flexible
  - block conflicts with internal entities and unintended uses
- reserve `mm-` as a prefix for names of containers, volumes, and ports in the runtime for future use
- if containers declare ports, check them against internal port usage to avoid conflicts (also reserve ports 11881-11899 for potential future use)

### DIFF
--- a/controllers/modelmesh/const.go
+++ b/controllers/modelmesh/const.go
@@ -14,8 +14,8 @@
 package modelmesh
 
 const (
-	ModelMeshContainer = "mm"
-	RESTProxyContainer = "rest-proxy"
+	ModelMeshContainerName = "mm"
+	RESTProxyContainerName = "rest-proxy"
 
 	GrpcPortEnvVar         = "INTERNAL_GRPC_PORT"
 	ServeGrpcPortEnvVar    = "INTERNAL_SERVING_GRPC_PORT"
@@ -31,7 +31,7 @@ const (
 	ConfigStorageMount = "storage-config"
 
 	//The name of the puller container
-	PullerContainer = "puller"
+	PullerContainerName = "puller"
 
 	//The env variable puller uses to configure it's own listen port
 	PullerEnvListenPort = "PORT"
--- a/controllers/modelmesh/const.go
+++ b/controllers/modelmesh/const.go
@@ -22,6 +22,12 @@ const (
 	GrpcUdsPathEnvVar      = "INTERNAL_GRPC_SOCKET_PATH"
 	ServeGrpcUdsPathEnvVar = "INTERNAL_SERVING_GRPC_SOCKET_PATH"
 
+	EtcdSecretKey = "etcd_connection"
+	EtcdVolume    = "etcd-config"
+
+	ModelsDirVolume = "models-dir"
+	SocketVolume    = "domain-socket"
+
 	ConfigStorageMount = "storage-config"
 
 	//The name of the puller container
--- a/controllers/servingruntime_validator.go
+++ b/controllers/servingruntime_validator.go
@@ -127,8 +127,8 @@ func checkName(name string, internalNames map[string]interface{}, logStr string)
 		return fmt.Errorf("%s %s is reserved for internal use", logStr, name)
 	}
 
-	if strings.HasPrefix(name, "mm") || strings.HasPrefix(name, "kserve") {
-		return fmt.Errorf("%s cannot start with \"mm\" or \"kserve\", which are reserved for internal use", logStr)
+	if strings.HasPrefix(name, "mm-") {
+		return fmt.Errorf("%s cannot start with \"mm-\", which is reserved for internal use", logStr)
 	}
 	return nil
 }
--- a/controllers/servingruntime_validator.go
+++ b/controllers/servingruntime_validator.go
@@ -80,11 +80,6 @@ func validateContainers(rt *api.ServingRuntime) error {
 
 func validateContainer(c *corev1.Container) error {
 	// Block container names that conflict with injected containers or reserved prefixes
-	internalContainerNames := []string{
-		modelmesh.ModelMeshContainerName,
-		modelmesh.RESTProxyContainerName,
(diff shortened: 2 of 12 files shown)


controllers/servingruntime_validator.go
```diff
@@ -0,0 +1,154 @@
...
+		if err := validateContainer(c); err != nil {
+			return fmt.Errorf("container spec for %s is invalid: %w", c.Name, err)
+		}
+	}
+	return nil
+}
+
+func validateContainer(c *corev1.Container) error {
+	// Block container names that conflict with injected containers or reserved prefixes
+	internalContainerNames := []string{
```
njhill, line 83: Maybe move this and the other sets of constants below outside of the functions?

Also could make these `map[string]interface{}`s with nil values, then the existence check is simpler.

controllers/servingruntime_validator.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := checkName(c.Name, internalContainerNames, "container name"); err != nil {
+		return err
+	}
+
+	// TODO: block if container name matches `*-adapter`??
+
+	// Block fields required for model-mesh to control the pod lifecycle
+	if c.ReadinessProbe != nil || c.Lifecycle != nil {
+		return errors.New("ReadinessProbe and Lifecycle are managed by modelmesh and cannot be set")
+	}
```
njhill, line 97: I'm not sure whether we should block these, at least not the liveness one. Though the desire is for model-mesh to own these and determine from the container itself, it might not do this well enough yet. I think some folks want to set their own.

njhill (commented):
Thanks @tjohnson31415, looks great

> - maybe should update names of internal volumes/containers to use the prefix

Sure, I have no preference on this really

> - should we reserve *-adapter for internal use or just check for conflict with the specific name generated for the "ServerType" (this would also be moot if the adapter container was prefixed with mm-, which is another option)

I'd say let's not bother with disallowing any -adapter suffix

> - should we reserve a port range for internal use?

I'd say no, the ports are only information anyhow right?

controllers/servingruntime_validator.go
tjohnson31415: The Liveness probe is not restricted here as that can be set in the current pre-Container SR spec. My goal here was to match the current behavior but maintain current restrictions where they make sense.

If you want the injected values for ReadinessProbe and Lifecycle to be overridable, I will remove this.

tjohnson31415:
@njhill 
> I'd say no, the ports are only information anyhow right?

In a sense, yeah, they are just informational (because they are optional / non-binding), but, if a Runtime provides that information, we can use it in these checks to try prevent bugs where a custom runtime attempts to use one of the internal ports used by ModelMesh.

The advantage to reserving a range of ports is to allow us to use an additional internal port in the future without worrying about breaking existing ServingRuntimes that may have a container using that port.

controllers/modelmesh/const.go
```diff
@@ -22,6 +22,12 @@ const (
 	GrpcUdsPathEnvVar      = "INTERNAL_GRPC_SOCKET_PATH"
 	ServeGrpcUdsPathEnvVar = "INTERNAL_SERVING_GRPC_SOCKET_PATH"
 
+	EtcdSecretKey = "etcd_connection"
+	EtcdVolume    = "etcd-config"
```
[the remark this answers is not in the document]
njhill: Ah ok, no that's fine, sorry I should have spotted it.

controllers/servingruntime_validator.go
njhill: OK yeah I mistakenly read this as checking both probes. I guess this is ok, we can always open up further if needed in future.

njhill:
@tjohnson31415 which ports would apply here? The port that modelmesh itself listens on is configurable - we could compare with the currently configured value but it would be kind of weird if updating that config suddenly caused existing SRs to stop working due to failed validation. The others I guess would be those used for rest-proxy -> modelmesh and modelmesh -> adapter communication? I don't remember the numbers we chose for these but I guess having them in some reserved range would make sense. So fine with me if you'd like to include that additional verification.

controllers/servingruntime_validator.go
```diff
@@ -0,0 +1,157 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func checkName(name string, internalNames map[string]interface{}, logStr string) error {
+	if internal, ok := internalNames[name]; ok {
+		return fmt.Errorf("%s %s is reserved for internal use", logStr, internal)
```
njhill, line 122:
```suggestion
	if _, ok := internalNames[name]; ok {
		return fmt.Errorf("%s %s is reserved for internal use", logStr, name)
```

tjohnson31415:
Pulling out of Draft.

I removed reserving `kserve-` for internal names as `kserve-container` [is the standard name](https://github.com/kserve/kserve/blob/3dc5bfb4a5d809819876f34ef926aca590438257/pkg/constants/constants.go#L240) for the inference container in a Runtime for KServe.

I also figure that this PR should focus on the CRD changes, so I left out renaming of internal containers/volumes/ports that may come later.

njhill (approved):
Thanks @tjohnson31415!

PR #152, opened 2022-04-29
test: Add keras FVT support

aluu317:
Added FVT support for Keras predictor


pvaneck (commented):
Thanks for adding this! I think for this case, the commits can just be squashed with the resulting commit signed-off so that we pass the DCO check. Then can add `Co-authored-by:` lines to the PR description so that any other authors can get credited.

pvaneck (approved):
Thanks, @aluu317!

/lgtm
/approve

PR #154, opened 2022-05-03
feat: Support for protocolVersion-based runtime matching

njhill:
#### Motivation

A `protocolVersions` list was [recently added](https://github.com/kserve/kserve/pull/2118) to the common `ServingRuntime` CRD, so that this can be taken into account when selecting a suitable runtime for a predictors that specify a protocol version.

#### Modifications

- Add `protocolVersion` field to `Predictor` CRD as well as conversion to/from corresponding `InferenceService` fields
- Update the existing controller logic that orchestrates the model-mesh type constraints configuration to take into account the ServingRuntime `protocolVersions` list in addition to the existing `supportedModelFormats` list
   - This also takes into account whether the REST proxy is enabled in the global config - if so, `V2` (REST) will be also included for any runtime declaring support for `GRPCV2`
- Update built-in runtimes and unit tests accordingly

#### Result

The runtime placement behaviour of model-mesh based predictors works as expected when they include a protocol version.

Supports https://github.com/kserve/kserve/issues/2065


njhill:
This is not fully functional yet since we first need to add the `protocolVersions` list to the `Predictor` CRD.

njhill:
Dev changes now complete but FVT failures need debugging.

tjohnson31415 (approved):
Overall LGTM. Left some minor NITs that you can incorporate or reject.

Having the REST proxy flag implicitly add supported protocol versions to the runtimes is a little awkward. Enabling the REST proxy in Runtimes could be an install time configuration instead of a runtime one. Something to consider improving in the future (not in this PR).

njhill:
Thanks for the review @tjohnson31415. I've addressed the comments, PTAL.

> Having the REST proxy flag implicitly add supported protocol versions to the runtimes is a little awkward.

I agree. Ideally we'll add regular HTTP support to modelmesh and eliminate the need for the proxy.

PR #155, opened 2022-05-03
refac: Replace custom StringSet impl with apimachinery sets.String

njhill:
#### Motivation

Our code included a simple `StringSet` implementation, but there's an equivalent type in the `k8s.io/apimachinery/pkg/util/sets` package.

#### Modifications

- Remove `StringSet` impl and associated unit tests
- Replace current use of `StringSet` with sets.String
- Use sets.String in `servingruntime_validator` rather than explicit maps
- Small fix to container port check in `servingruntime_validator`

#### Result

Less/clearer code, better reuse

### DIFF
--- a/controllers/servingruntime_validator.go
+++ b/controllers/servingruntime_validator.go
@@ -17,6 +17,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"errors"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


controllers/servingruntime_validator.go
```diff
@@ -17,6 +17,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
tjohnson31415, line 20: NIT: the spacing in this import block is strange; I think `"errors"` should be up with the other stdlib packages and this down with the other `k8s.io` import.

njhill: @tjohnson31415 updated, PTAL

tjohnson31415:
PR description should be updated after the change to use sets.Int32, but the code

/lgtm

PR #157, opened 2022-05-05
chore: convert quickstart/fvt pods into deployments

pvaneck:
The minio and etcd instances now use deployments instead of pods to provide
some level of resiliency. 

The quickstart doc was also updated with a small disclaimer about these deployments.


njhill (approved):
LGTM thanks @pvaneck

PR #160, opened 2022-05-20
fix: Fix possible nil pointer dereference; minor log improvements

njhill:
#### Motivation

During unrelated testing I found a bug where a nil pointer deref could occur if ServingRuntimes are encountered in namespaces in which modelmesh isn't enabled (i.e. not the controller's namespace and without the `modelmesh-enabled=true` label).

#### Modifications

- Move list pointer initialization to ensure it won't be nil
- Improve some badly worded log statements and one containing excessive detail (full deployment contents rather than just name)


chinhuang007 (approved):
/lgtm

Issue #136, opened 2022-03-17
Add ModelMetadata endpoint support


pvaneck:
Users may want to find the input tensor name and/or shape of a particular model. The v2 predict protocol outlines a [ModelMetadata API](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#model-metadata-1). We should have ModelMesh support this as I believe Triton and MLServer already expose this endpoint.

Related issue and discussion: https://github.com/kserve/modelmesh-serving/issues/104

njhill:
I think all that should be needed is the addition of that rpc to those advertised by the adapter logic for Triton and MLServer [here](https://github.com/kserve/modelmesh-runtime-adapter/blob/main/model-mesh-triton-adapter/server/server.go#L296) and [here](https://github.com/kserve/modelmesh-runtime-adapter/blob/main/model-mesh-mlserver-adapter/server/server.go#L560), as well as to the id extraction config set by the controller [here](https://github.com/kserve/modelmesh-serving/blob/c8f9cf129a2f231df1fb75bdc858b21d430df84d/controllers/modelmesh/cluster_config.go#L36). And test coverage of course.

PR #164, opened 2022-06-09
feat: Enable ModelMetadata endpoint

pvaneck:
**Motivation**

Enable users to use the KServe v2 ModelMetadata endpoint with Triton and MLServer ServingRuntimes.

**Modifications**

- ModelMetadata added to dataplane config in controller.
- Some functional tests were added to test the newly exposed endpoint.

**Result**

Users can send gRPC requests to the ModelMetadata endpoint. Primarily useful with Triton as MLServer only seems to return the name of the model.

Partially-Closes: #136

### DIFF
--- a/fvt/fvtclient.go
+++ b/fvt/fvtclient.go
@@ -425,6 +425,18 @@ func (fvt *FVTClient) RunKfsInference(req *inference.ModelInferRequest) (*infere
 	return grpcClient.ModelInfer(ctx, req)
 }
 
+func (fvt *FVTClient) RunKfsModelMetadata(req *inference.ModelMetadataRequest) (*inference.ModelMetadataResponse, error) {
+	if fvt.grpcConn == nil {
+		return nil, errors.New("you must connect to model mesh before running a model metadata request")
+	}
+
+	grpcClient := inference.NewGRPCInferenceServiceClient(fvt.grpcConn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	return grpcClient.ModelMetadata(ctx, req)
+}
+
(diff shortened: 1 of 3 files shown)


config/default/config-defaults.yaml
```diff
@@ -38,7 +38,7 @@ restProxy:
       cpu: "1"
       memory: "512Mi"
 storageHelperImage:
-  name: kserve/modelmesh-runtime-adapter
+  name: pvaneckw/modelmesh-runtime-adapter
```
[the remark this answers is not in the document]
pvaneck: Yea, I see that it was merged (thanks!). I will switch back the image now.

njhill (approved):
Thanks @pvaneck!

fvt/fvtclient.go
```diff
@@ -425,6 +425,18 @@ func (fvt *FVTClient) RunKfsInference(req *inference.ModelInferRequest) (*infere
 	return grpcClient.ModelInfer(ctx, req)
 }
 
+func (fvt *FVTClient) RunKsModelMetadata(req *inference.ModelMetadataRequest) (*inference.ModelMetadataResponse, error) {
```
chinhuang007, line 428: Nit: maybe RunKfsModelMetadata to be consistent?

pvaneck: Initially wanted to start using the abbreviation for the rename, but, sure, I'll keep it consistent for now. Maybe at some point, we can do a mass rename.

chinhuang007: Yeah, I thought you might have ks for KServe to be different from ksf. Agreed, at some point, we should rename all kfs.

chinhuang007 (approved):
Thanks, @pvaneck !

PR #166, opened 2022-06-13
chore: Include InferenceService CRD in standalone mm-serving install

njhill:
#### Motivation

Currently modelmesh-serving does not include the `InferenceService` CRD manifest, but will work with it if it's been previously installed (via an install of the main kserve controller). Since we will be encouraging everyone to use this in preference to the `Predictor` CRD, it would be good if it was available by default in standalone modelmesh installations.

#### Modifications

- Copy manifests of common CRDs including inferenceservices from https://github.com/kserve/kserve/tree/master/config/crd
- No longer pull from kserve for fvt tests
- Update make manifests target to include v1beta1

#### Result

`InferenceService` CRD will be available to use when using modelmesh-serving "standalone" without having to obtain it manually from the kserve/kserve repo.

### DIFF
--- a/.github/workflows/run-fvt.yml
+++ b/.github/workflows/run-fvt.yml
@@ -56,7 +56,6 @@ jobs:
       - name: Install ModelMesh Serving
         run: |
           kubectl create ns modelmesh-serving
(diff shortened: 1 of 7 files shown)


njhill:
Thanks @pvaneck actually that file _is_ needed for the controller, I added it back and updated the commit message (but forgot to update the PR description, now updated too).

PR #167, opened 2022-06-15
docs: Add docs for OpenVINO

aluu317:
Signed-off-by: Angel Luu <angel.luu@us.ibm.com>

#### Motivation

Update docs after the changes for OVMS and `openvino-ir` model support in #141 .

### DIFF
--- a/docs/model-types/README.md
+++ b/docs/model-types/README.md
@@ -21,6 +21,6 @@ By leveraging existing third-party model servers, we support a number of standar
 | sklearn     | scikit-learn     | MLServer (python)            |
 | tensorflow  | TensorFlow       | Triton (C++)                 |
 | xgboost     | XGBoost          | MLServer (python)            |
-| \*\*        | Custom           | [Custom](../runtimes) (any)  |
+| any         | Custom           | [Custom](../runtimes) (any)  |
 
 (\*)Many ML frameworks can have models converted to the OpenVINO IR format, such as Caffe, TensorFlow, MXNet, PaddlePaddle and ONNX, doc [here](https://docs.openvino.ai/latest/ovms_what_is_openvino_model_server.html).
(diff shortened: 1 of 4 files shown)


docs/model-types/README.md
```diff
@@ -5,18 +5,22 @@ By leveraging existing third-party model servers, we support a number of standar
...
+| keras       | TensorFlow       | Triton (C++)                 |
+| lightgbm    | LightGBM         | MLServer (python)            |
+| onnx        | ONNX             | Triton (C++), OVMS (C++)     |
+| openvino_ir | Intel OpenVINO\* | OVMS (C++)                   |
+| pytorch     | PyTorch          | Triton (C++)                 |
+| sklearn     | scikit-learn     | MLServer (python)            |
+| tensorflow  | TensorFlow       | Triton (C++)                 |
+| xgboost     | XGBoost          | MLServer (python)            |
+| \*\*        | Custom           | [Custom](../runtimes) (any)  |
+
```
chinhuang007, line 25: Looks good, thanks!
I thought the footnote is mainly for the new built-in OVMS runtime. Not sure the use of double asterisk symbols for the row that is for custom runtimes. In fact, models in many ml frameworks can be converted to ONNX IR and run in Triton or OVMS. I would recommend remove the second asterisk.

njhill: I think the asterisk in custom row originally indicated "any", so I agree it doesn't make sense to also use an asterisk for a footnote indicator in the same table. Maybe use a superscript `1` if that's possible in markdown, or otherwise some other symbol.

chinhuang007 (approved):
/lgtm

docs/model-types/README.md
chinhuang007: Looks good. Thanks!

PR #172, opened 2022-06-23
chore: move common apis back into v1alpha1

anhuong:
#### Motivation

Move common apis back to v1alpha1 since they are no longer "common" after changed modelmesh-serving to use the inferenceservice types from kserve. Most of those types are now in the kserve repo inferenceservice types file.

#### Modifications

- Move common apis back to v1alpha1 predictor_types (where they were initially)
- Regenerate manifests using controller-gen v0.4.0
   - Mostly add `descriptions:` to ClusterServingRuntime and Runtime
   - Lots of changes to InferenceService Also added some `require:` fields (port)
   - `default: TCP` deleted from 

#### Result

Better align with kserve for apis


anhuong:
Should I not change `serving.kserve.io_inferenceservices.yaml` and continue to pull it from https://github.com/kserve/kserve/blob/400078ed3e98362dc434e4ece0d0d82678dd9f0b/config/crd/serving.kserve.io_inferenceservices.yaml ?

pvaneck (commented):
Thanks, @anhuong!

For this PR, let's revert all changes to the configs as it doesn't look like the `serving.kserve.io_predictors.yaml` changed at all, and the other config changes aren't related to the changes in common_types.

The ServingRuntime, ClusterServingRuntime, and InferenceService CRDs are now essentially just mirrored from KServe (maybe at some point we should remove the generation of these from `make manifests`).

pvaneck (commented):
Thanks! 
/lgtm
/approve

PR #174, opened 2022-06-24
feat: Support azure blob storageURIs in Isvc

pvaneck:
#### Motivation

The model puller now supports pulling from Azure Blob Storage. We should allow users to pass in an
Azure Blob Storage URI as an InferenceService Predictor `storageUri`.

#### Modifications

Storage parsing logic now parses Azure Blob URIs and passes the relevant info the puller.

#### Result

Users can now pass in Azure Blob Storage URIs as part of `storageUri`.


pvaneck:
Note: this was tested with public and private blob containers in azure, and was able to get the Isvc up and running.

chinhuang007 (approved):
/lgtm

Issue #173, opened 2022-06-23
ProtocolVersion required when deploying Isvc with old format


pvaneck:
This works:

```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/deploymentMode: ModelMesh
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    model:
      modelFormat:
        name: sklearn
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```

while: 

```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: example-sklearn-isvc
  annotations:
    serving.kserve.io/deploymentMode: ModelMesh
    serving.kserve.io/secretKey: localMinIO
spec:
  predictor:
    sklearn:
      storageUri: s3://modelmesh-example-models/sklearn/mnist-svm.joblib
```

this does not. 

In the second case, `getRuntimesSupportingPredictor()` isn't returning anything unless something like `protocolVersion: grpc-v2` is explicitly specified. 

The old format should still be able to find a runtime like in the first case even if no protocolVersion is specified by the user.

PR #178, opened 2022-06-29
fix: Ensure correct protocolVersion is retrieved

pvaneck:
For InferenceServices using the old predictor schema, no
matching runtime is found because GetProtocol generally
returns a protocol even if none is specified in the spec.

This adjusts the logic a bit to pull the ProtocolVersion from the actual spec.

Fixes: #173


chinhuang007 (approved):
Looks good! Thanks, @pvaneck !

chinhuang007:
Verified with Kserve including the webhook fix.

PR #181, opened 2022-07-06
chore: Rename roles/rolebindings

pvaneck:
#### Motivation

The role/rolebinding names should be more specific to enhance clarity of purpose
and to avoid potential naming collisions.

#### Modifications

Prepended a `modelmesh-` prefix to relevant names. 

#### Result

More clarity for what each role/rolebindings belongs to.


chinhuang007 (approved):
Looks good. Thanks, @pvaneck !

PR #187, opened 2022-07-13
Update performance test script for e2e pipeline

chinhuang007:
Update performance test script to deploy the predictor first
to ensure it is up and runnig before k6 is executed during
pipeline runs.

It takes about 20 seconds in test environments so the retry
loop will last for 60 seconds to be safe.


pvaneck (commented):
/lgtm
/approve

Issue #186, opened 2022-07-13
The etcd connection is incorrect for user namespaces when etcd is configured outside of controller namespace


chinhuang007:
**Describe the bug**

If etcd is configured outside of the controller namespace, the generated secret for user namespaces will have an incorrect etcd endpoint.

**To Reproduce**
1. Configure etcd in a separate namespace
2. Deploy modelmesh including a controller namespace, using the etcd created in step 1
3. Enable modelmesh in a user namespace
4. Create a ServingRuntime, which won't work due to the incorrect etcd endpoint 

**Expected behavior**

The generate secret should contain correct etcd endpoints when etcd is in a different namespace from controller.

XiangFrank:
Hi, we found another issue regarding Etcd secret copy. When copying the Etcd secret, seems it would only copy contents in etcd_connection field rather than the whole secret. Is there any reason behind it? Because we're using client_key_file to pass the key in another field, and it was skipped during the copying process.

chinhuang007:
> Is it possible to directly add the namespace information in the Etcd endpoint at the beginning if it is exposed as an internal service? In this case, the extra logic in addData could be removed, and even if users are directly using external IP addresses as the Etcd endpoints it could still be safe.

Thanks for the suggestion. It is what I was thinking as well. Feel free to check out and comment on the work-in-progress patch, https://github.com/kserve/modelmesh-serving/pull/188

pvaneck:
@XiangFrank The `etcd_connection` key in the secret supports other fields like `client_key_file` as shown [here](https://github.com/kserve/modelmesh-serving/blob/main/pkg/mmesh/etcdrangewatcher.go#L32). Is your `client_key_file` being set as a separate key of the secret?

PR #188, opened 2022-07-13
Fix etcd endpoints when etcd is in different namespace

chinhuang007:
Fix the etcd endpoints issue when etcd is in different namespace
as described in https://github.com/kserve/modelmesh-serving/issues/186

The default etcd connection endpoints in quick start and fvt yaml are
changed to be placceholder based, and will be replaced during the
install script run. The etcd secret propagation will no longer assume
that etcd resides in the controller namespace.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -225,19 +225,23 @@ kustomize edit set namespace "$namespace"
 cd ..
 
 # Clean up previous instances but do not fail if they do not exist
+cp dependencies/quickstart.yaml .
+sed -i.bak "s/etcd:2379/etcd.${namespace}:2379/g" quickstart.yaml
+cp dependencies/fvt.yaml .
+sed -i.bak "s/etcd:2379/etcd.${namespace}:2379/g" fvt.yaml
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -225,19 +225,23 @@ kustomize edit set namespace "$namespace"
 cd ..
 
 # Clean up previous instances but do not fail if they do not exist
+cp dependencies/quickstart.yaml .
+sed -i.bak "s/etcd:2379/etcd.${namespace}:2379/g" quickstart.yaml
```
yuzisun, line 229: you can use `sed -pi -e` instead of generating the bak file.

chinhuang007: sed is implemented differently in OSX and Linux (GNU). According to the [spec](https://pubs.opengroup.org/onlinepubs/9699919799/), -p is not standard. `sed -pi -e` actually fails in Ubuntu/GNU sed. The use of bak works for both GNU and BSD versions of sed.

pvaneck (approved):
/lgtm
/approve

PR #202, opened 2022-08-01
feat: Ensure modelmesh container comes last in list

njhill:
#### Motivation

Kubernetes starts containers sequentially in order, which can mean the start of later containers is held up if their images have to be pulled. For large model server images this can cause a problem because model-mesh waits for a limited amount of time at startup for the runtime to become ready before returning ready from its own probe.

Making the `mm` container last in the list ensures that no image pulling time will be included in this startup polling time and avoids unnecessary timeouts / extended readiness probe failures.

#### Modifications

- Add `ensureMMContainerIsLast` function to runtime deployment reconciliation logic which runs after other modifications to the pod spec.
- Update unit test fixtures accordingly

#### Result

More stable deployments in cases where the model server image takes a long time to pull.


chinhuang007 (approved):
Looks good!

PR #212, opened 2022-08-19
fix: Fix namespace-scoped install

njhill:
#### Motivation

There are some problems when installing in namespace-scoped mode:
- Though the docs mentioned that the RBAC should be changed, the install script currently applies the controller `ClusterRole` and `ClusterRoleBinding` (including namespace resource permission) regardless of whether the `--namespace-scope-mode` option is provided
- The combination of the `NAMESPACE_SCOPE=true` env var and namespace resource permissions results in a broken/conflicted state where the controller will "mostly" operate in cluster scoped mode and watch / delete resources in other namespaces

#### Modifications
- Split RBAC manifests into cluster-scope and namespace-scope variants, the latter including `Role`/`RoleBinding` instead of Cluster level equivalents
- Have install script apply the appropriate RBAC based on the namespace-scope-mode cli setting (in addition to setting the controller env var as is already done)
- Change the controller init logic to always use namespace mode if the env var is set
- Update documentation accordingly

#### Result

Namespace scoped install works properly, won't touch any other namespaces.

It would probably also be good to add some safeguard checks to detect/prohibit cluster scope installs in more than one namespace, but that's outside of the scope of this PR.


scripts/delete.sh
```diff
@@ -115,7 +118,17 @@ if [[ ! -z $user_ns_array ]]; then
   rm runtimes.yaml
 fi
 
+# Determine whether a modelmesh-controller-rolebinding clusterrolebinding exists and is
+# associated with the service account in this namespace. If not, don't delete the cluster level RBAC.
+set +e
+crb_ns=$(oc get clusterrolebinding modelmesh-controller-rolebinding -o json | jq -r .subjects[0].namespace)
```
chinhuang007, line 124: Shall we stay with kubectl instead of oc?

njhill: Thanks @chinhuang007 good catch, now fixed.

chinhuang007 (approved):
Looks good, thanks, @njhill !

PR #220, opened 2022-08-29
feat: Make list of supported built-in runtime server types overridable

njhill:
#### Motivation

Model-mesh serving makes use of a single adapter image to interface with a number of "built-in" model server types, currently comprising `triton`, `mlserver`, `ovms`.

Currently which kind of adapter to run is controlled by the entrypoint to the adapter container within the modelmesh pods, which is set by the controller based on the `builtInAdapter.serverType` field of the `ServingRuntime` CRD.

However this is currently validated against a hardcoded list of type strings meaning that any extensions to the runtime adapter image to support new kinds of model servers also require a code change and rebuild of the controller image.

#### Modifications

Move the list of supported built-in server types to a list in the global config.
Update controller validation logic and tests accordingly.

#### Result

The shared built-in runtime adapter can be extended to support new runtime types without requiring a code change to the controller.

### DIFF
--- a/config/default/config-defaults.yaml
+++ b/config/default/config-defaults.yaml
@@ -51,3 +51,7 @@ storageHelperResources:
 serviceAccountName: ""
 metrics:
   enabled: true
+builtInServerTypes:
+  - triton
+  - mlserver
+  - ovms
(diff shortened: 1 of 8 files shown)


njhill:
@chinhuang007 I'm not sure why this is failing the FVTs. I have a feeling it may be a resource constraint thing but I can't see how the changes in the PR could have affected that.

chinhuang007 (approved):
Looks good, thanks, @njhill !

PR #227, opened 2022-09-07
feat: Pass ServingRuntime inference endpoint to built-in adapter

njhill:
#### Motivation

It may be useful for some built-in runtime adapters to have the model server's inferencing endpoint information in addition to the existing "management" port number that's passed (in case it is different).

#### Modifications

- Set a new `RUNTIME_DATA_ENDPOINT` env var on the built-in adapter container
- Parse the `ADAPTER_PORT` env var value from the ServingRuntime `grpcEndpoint` field instead of hardcoding to 8085

### DIFF
--- a/controllers/modelmesh/runtime.go
+++ b/controllers/modelmesh/runtime.go
@@ -15,6 +15,7 @@ package modelmesh
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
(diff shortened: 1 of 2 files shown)


chinhuang007 (approved):
Looks good, thanks, @njhill !

PR #235, opened 2022-09-08
doc: Update FVT prerequisites

ckadner:
#### Motivation

After many attempts to run the FVT tests it became clear that an undersized Kubernetes cluster (insufficient resources) prevented the FVTs to succeed.

#### Modifications

Update the FVT README with a Prerequisite section that outlines the required CLIs and the minimum cluster size.

#### Result

The FVT tests completed successfully.

/cc @njhill @chinhuang007

### DIFF
--- a/fvt/README.md
+++ b/fvt/README.md
@@ -10,6 +10,19 @@ Functional Verification Test (FVT) suite for ModelMesh Serving using [Ginkgo](ht
 
 ## How to run the FVT suite
 
+### Prerequisites
+
+- CLIs:
+  - [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl)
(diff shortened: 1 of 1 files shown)


njhill:
@ckadner there's a linter failure, you need to run `make fmt` which will fail the first time but make the required formatting changes. So running a second time should pass and you can then commit those changes.

ckadner:
> @ckadner there's a linter failure, you need to run `make fmt` which will fail the first time but make the required formatting changes. So running a second time should pass and you can then commit those changes.

Thanks @njhill -- I am too used to my IDE doing the code cleanup for me

ckadner:
Thanks **Nick**!

Issue #81, opened 2021-12-03
ClusterServingRuntime support


pvaneck:
KServe supports cluster-scoped ServingRuntimes called `ClusterServingRuntimes`. These act as the built-in or default serving runtimes accessible to any user/namespace in the cluster. Currently ModelMesh-Serving only considers the the namespace-scoped `ServingRunimes`. Let's think about how ModelMesh-Serving can handle these cluster-level resources.

njhill:
Summary of design discussion with @chinhuang007:
- A `ServingRuntime` in a given namespace will "hide" a `ClusterServingRuntime` with the same name for any inference services in that namespace
- Deployments derived from `ClsuterServingRuntime`s won't have an owner set, otherwise the `ServingRuntime` will be set as the owner as it is now
- When installing in cluster scope mode (default), the built-in runtimes will be created as `ClusterServingRuntime`s, otherwise they will be created as regular `ServingRuntime`s in the target namespace

Code changes required:
- Have the SR reconciler also watch CSRs, with a handler function that maps CSR events to multiple SR requests, one per "modelmesh-enabled" namespace
- In the Reconcile function, always `Get` the SR as we do now. If it is not found, attempt to `Get` a CSR with the same name
- Refactor the rest of Reconcile function to work with `*ServingRuntimeSpec` (which exists in both CR and CSR) rather than SR directly
- Where list of runtimes is retrieved, instead create merged list of SR and CSR with SRs taking precedence

We should have the CSR parts conditional on some flag so that things will work the same as they do now if that CRD does not exist (or controller does not have permisison to list/read them). We may also want to expose and option to explicitly enable/disable use of CSRs in namespace-scope case.

chinhuang007:
I have shared the discussions, design, and implement [here](https://docs.google.com/document/d/1lSqwqmiOeS7rJTtxfSdvxuTK_GzaOwTV5BXfnw2IEJs/edit?usp=sharing). It is a bit more elaborated than what @njhill summarized above. Feel free to make comments!

njhill:
Required changes to the controller have now been completed in #241. Remaining tasks:
- [ ] Update config/install scripts to install CSR CRD and built-in runtimes as CSRs for cluster-scope mode
- [ ] Ensure FVT coverage of both SRs and CSRs

rafvasq:
Closed by https://github.com/kserve/modelmesh-serving/pull/241.

[...]

PR #241, opened 2022-09-13
Add ClusterServingRuntime support

chinhuang007:
Add ClusterServingRuntime (CSR) support so that the CSRs can be shared between namespaces and overridden by namespaece level ServingRuntimes (SR).

#### Motivation
This PR provides the support for ClusterServingRuntime (CSR), addressing [issue ](https://github.com/kserve/modelmesh-serving/issues/81). The use cases and design can be found [here](https://docs.google.com/document/d/1lSqwqmiOeS7rJTtxfSdvxuTK_GzaOwTV5BXfnw2IEJs/edit#heading=h.x1r6y17xd8u7).

#### Modifications

- ServingRuntime controller watches CSRs
- ServingRuntime controller determines which CSRs can be used to create deployments in ModelMesh enabled namespaces
- ServingRuntime controller creates deployments using CSRs
- ServingRuntime controller decides whether a SR or a CSR to use for the deployment
- ServingRuntime controller sets deployment owner=nil for CSRs
- Refactor code to take SR.spec instead of SR and use the spec from either a SR or a CSR.
- Change the install script/config to install CSRs in the cluster scope mode.

#### Result
ModelMesh supports ClusterServingRuntimes while namespace level ServingRuntimes have overriding power.


controllers/servingruntime_controller.go
```diff
@@ -166,35 +176,62 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// SR is found
+		rt2 = rt
+	} else {
+		// CSR is found
+		spec = &crt.Spec
+		rt2 = nil
+		rtName = types.NamespacedName{Name: req.Name}
+	}
+	log.Info("==== ChinDebug sr reconcile 2.5 ====", "spec", spec)
+	log.Info("==== ChinDebug sr reconcile 2.5 ====", "rtName", rtName)
```
njhill, line 218: This logic is the right idea but can be simplified I think, e.g:
```go
	var spec *kserveapi.ServingRuntimeSpec
	rt := &kserveapi.ServingRuntime{}
	if err = r.Client.Get(ctx, req.NamespacedName, rt); err == nil {
		spec = &rt.Spec
	} else if errors.IsNotFound(err) {
		log.Info("Runtime is not found in namespace")

		// try to find the runtime in cluster ServingRuntimes
		crt := &kserveapi.ClusterServingRuntime{}
		if err = r.Client.Get(ctx, types.NamespacedName{Name: req.Name}, crt); err == nil {
			spec = &crt.Spec
			rt = nil
		} else if errors.IsNotFound(err) {
			log.Info("Runtime is not found in cluster")

			// remove runtime from info map
			r.runtimeInfoMapMutex.Lock()
			defer r.runtimeInfoMapMutex.Unlock()

			if r.runtimeInfoMap != nil {
				// this is safe even if the entry doesn't exist
				delete(r.runtimeInfoMap, req.NamespacedName)
			}
			return ctrl.Result{}, nil
		} else {
			return ctrl.Result{}, fmt.Errorf("error retrieving ClusterServingRuntime %s: %w", req.Name, err)
		}
	} else {
		return ctrl.Result{}, fmt.Errorf("error retrieving ServingRuntime %s: %w", req.NamespacedName, err)
	}
```

controllers/servingruntime_controller.go
```diff
@@ -222,21 +259,25 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		AnnotationsMap:      cfg.RuntimePodAnnotations,
 		LabelsMap:           cfg.RuntimePodLabels,
 	}
+	log.Info("==== ChinDebug sr reconcile 5 ==== before delete mmDeployment", "mmEnabled", mmEnabled)
 
 	// if the runtime is disabled, delete the deployment
-	if rt.Spec.IsDisabled() || !rt.Spec.IsMultiModelRuntime() || !mmEnabled {
+	if (*spec).IsDisabled() || !(*spec).IsMultiModelRuntime() || !mmEnabled {
```
njhill, line 265:
```suggestion
	if spec.IsDisabled() || !spec.IsMultiModelRuntime() || !mmEnabled {
```

controllers/servingruntime_controller.go
```diff
@@ -166,35 +176,62 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	log.Info("==== ChinDebug sr reconcile 3 ==== before mmDeployment", "cfg", cfg)
+
 	// construct the deployment
 	mmDeployment := modelmesh.Deployment{
 		ServiceName:                cfg.InferenceServiceName,
 		ServicePort:                cfg.InferenceServicePort,
 		Name:                       req.Name,
 		Namespace:                  req.Namespace,
-		Owner:                      rt,
+		Owner:                      rt2,
```
njhill, line 233: This can remain as `rt` with the above simplification.

controllers/servingruntime_controller.go
```diff
@@ -222,21 +259,25 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		log.Info("Runtime is disabled, incompatible with modelmesh, or namespace is not modelmesh-enabled")
 		if err = mmDeployment.Delete(ctx, r.Client); err != nil {
 			return ctrl.Result{}, fmt.Errorf("could not delete the model mesh deployment: %w", err)
 		}
 		return ctrl.Result{}, nil
 	}
+	log.Info("==== ChinDebug sr reconcile 6 ==== no need to delete mmDeployment", "mmEnabled", mmEnabled)
 
-	replicas, requeueDuration, err := r.determineReplicasAndRequeueDuration(ctx, log, cfg, rt)
+	replicas, requeueDuration, err := r.determineReplicasAndRequeueDuration(ctx, log, cfg, spec, rtName)
```
njhill, line 274:
```suggestion
	replicas, requeueDuration, err := r.determineReplicasAndRequeueDuration(ctx, log, cfg, spec, req.NamespacedName)
```
(no need for intermediate var)

controllers/servingruntime_controller.go
```diff
@@ -369,23 +412,44 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
...
+	if err := r.Client.List(ctx, csrs); err != nil {
+		return nil, err
+	}
+
 	restProxyEnabled := r.ConfigProvider.GetConfig().RESTProxy.Enabled
 	srnns := make([]types.NamespacedName, 0, len(runtimes.Items))
 	for i := range runtimes.Items {
 		rt := &runtimes.Items[i]
-		if rt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(rt, p, restProxyEnabled) {
+		if rt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(&rt.Spec, p, restProxyEnabled, types.NamespacedName{Name: rt.Name, Namespace: rt.Namespace}) {
```
njhill, line 425: Per other comment only need to pass string `rtName` here, not NamespacedName

controllers/servingruntime_controller.go
```diff
@@ -369,23 +412,44 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
...
 			srnns = append(srnns, types.NamespacedName{
 				Name:      rt.GetName(),
 				Namespace: p.Namespace,
 			})
 		}
 	}
+	csrnns := make([]types.NamespacedName, 0, len(csrs.Items))
```
njhill, line 432: no need for a separate slice here, can just append to the existing one

controllers/servingruntime_controller.go
```diff
@@ -369,23 +412,44 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
...
+		crt := &csrs.Items[i]
+		// Check whether a ServingRuntime exists with same name
+		if crt.Spec.IsMultiModelRuntime() && !r.runtimeExists(runtimes, crt.Name) &&
+			runtimeSupportsPredictor(&crt.Spec, p, restProxyEnabled, types.NamespacedName{Name: crt.Name}) {
+			csrnns = append(csrnns, types.NamespacedName{
+				Name:      crt.GetName(),
+				Namespace: p.Namespace,
+			})
+		}
+	}
```
njhill, line 443: We could probably simplify the list merging a bit, e.g. by populating a map instead of a slice (then an explicit runtimeExists check won't be needed)

controllers/servingruntime_controller.go
```diff
@@ -480,3 +553,50 @@ func (r *ServingRuntimeReconciler) runtimeRequestsForPredictor(p *api.Predictor,
 	}
 	return requests
 }
+
+func (r *ServingRuntimeReconciler) clusterServingRuntimeRequests(csr *kserveapi.ClusterServingRuntime, source string) []reconcile.Request {
```
njhill, line 557: `source` param not needed here

controllers/servingruntime_controller.go
```diff
@@ -480,3 +553,50 @@ func (r *ServingRuntimeReconciler) runtimeRequestsForPredictor(p *api.Predictor,
...
+	var ToReconcile = csr.Spec.MultiModel != nil && *csr.Spec.MultiModel
+	r.Log.Info("=== ChinDebug process CSR ToReconcile ===", "ToReconcile", ToReconcile)
+
+	if ToReconcile {
+		// get list of namespaces
+		var opts []client.ListOption
+		list := &corev1.NamespaceList{}
+
+		// return nothing if can't get namespaces
+		if err := r.Client.List(context.TODO(), list, opts...); err != nil || len(list.Items) == 0 {
```
njhill, line 569: No need for `opts` var, can just exclude that function param

controllers/servingruntime_controller.go
```diff
@@ -480,3 +553,50 @@ func (r *ServingRuntimeReconciler) runtimeRequestsForPredictor(p *api.Predictor,
 	}
 	return requests
 }
+
+func (r *ServingRuntimeReconciler) clusterServingRuntimeRequests(csr *kserveapi.ClusterServingRuntime, source string) []reconcile.Request {
+	r.Log.Info("=== ChinDebug process CSR name ===", "csr", csr.Name)
+
+	var ToReconcile = csr.Spec.MultiModel != nil && *csr.Spec.MultiModel
```
njhill, line 560: Would be a bit simpler to just return here in non-multimodel case, i.e.
```go
if csr.Spec.MultiModel == nil || !*csr.Spec.MultiModel {
    return []reconcile.Request{}
}
```

config/runtimes/cluster-scope/mlserver-0.x.yaml
```diff
@@ -0,0 +1,75 @@
+# Copyright 2021 IBM Corporation
```
njhill, line 1: I think using kustomize we can avoid new yaml files. We can change the existing ones to be `ClusterServingRuntime`s and then for namespace scope install have an overlay which patches them like:

```yaml
patchesJSON6902:
- target:
    group: serving.kserve.io
    version: v1alpha1
    kind: ClusterServingRuntime
    name: ".*"
  patch: |-
    - op: replace
      path: /kind
      value: ServingRuntime
```

controllers/modelmesh/modelmesh.go
```diff
@@ -281,6 +314,9 @@ func (m *Deployment) addMMEnvVars(deployment *appsv1.Deployment) error {
 }
 
 func (m *Deployment) setConfigMap() error {
+	if m.Owner == nil {
+		return nil
+	}
```
njhill, line 319: This can use annotations stored in the Deployment struct (see other comment about that).

njhill (commented):
Hey thanks @chinhuang007 this is a great start and not far off I think.

Something additional I didn't include inline comments about: In a few places the CSR logic should be skipped if CSR watching is disabled.

controllers/modelmesh/runtime.go
```diff
@@ -68,7 +68,7 @@ func (m *Deployment) addVolumesToDeployment(deployment *appsv1.Deployment) error
 	}
 
 	// need to mount storage config for built-in adapters and the scenarios where StorageHelper is not disabled
-	if rt.Spec.BuiltInAdapter != nil || useStorageHelper(rt) {
+	if m.SRSpec.BuiltInAdapter != nil || useStorageHelper(rts) {
```
njhill, line 71:
```suggestion
	if rts.BuiltInAdapter != nil || useStorageHelper(rts) {
```

chinhuang007:
@njhill Thanks for your thorough review! Most of your comments are addressed. Still look into the install/kustomization comment and will remove debugging in a separate commit later.

controllers/servingruntime_controller.go
```diff
@@ -166,26 +192,42 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		// remove runtime from info map
-		r.runtimeInfoMapMutex.Lock()
-		defer r.runtimeInfoMapMutex.Unlock()
-
-		if r.runtimeInfoMap != nil {
-			// this is safe even if the entry doesn't exist
-			delete(r.runtimeInfoMap, req.NamespacedName)
+	crt := &kserveapi.ClusterServingRuntime{}
+	var spec *kserveapi.ServingRuntimeSpec
+	var annotations map[string]string
```
njhill, line 199: Per other comment I think we can remove annotations here, and instead of `crt` var at this level can have an `mf.Owner` var which is set to one or the other below (where annotations are currently set).

njhill (commented):
Thanks @chinhuang007 I'm still reviewing the latest changes but here's a couple of initial comments.

controllers/servingruntime_controller.go
```diff
@@ -437,32 +429,31 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
 	}
 
 	restProxyEnabled := r.ConfigProvider.GetConfig().RESTProxy.Enabled
-	srnns := make([]types.NamespacedName, 0, len(runtimes.Items))
+	srNameMap := make(map[string]bool)
```
njhill, line 432: nit: It's actually slightly smaller to make a set with `map[string]struct{}`.. and to insert use `... = struct{}{}`

controllers/servingruntime_controller.go
```diff
@@ -437,32 +429,31 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
...
+	for i := range csrs.Items {
+		crt := &csrs.Items[i]
+		if crt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(&crt.Spec, p, restProxyEnabled, crt.Name) {
```
njhill, line 435: Maybe move this check from these two loops to the single loop below? Then it only needs to be in one place

controllers/servingruntime_controller.go
```diff
@@ -369,23 +420,45 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
...
-		if rt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(rt, p, restProxyEnabled) {
-			srnns = append(srnns, types.NamespacedName{
-				Name:      rt.GetName(),
-				Namespace: p.Namespace,
-			})
+		if rt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(&rt.Spec, p, restProxyEnabled, rt.Name) {
+			srNameMap[rt.GetName()] = true
 		}
 	}
+	srnns := make([]types.NamespacedName, 0, len(srNameMap))
```
njhill, line 445: We could consider changing this function to return the map rather than building a slice here, since the calling function just builds another slice.

main.go
```diff
@@ -394,7 +421,7 @@ func main() {
 		ControllerName:      controllerDeploymentName,
 		HasNamespaceAccess:  clusterScopeMode,
 		RegistryMap:         registryMap,
-	}).SetupWithManager(mgr, enableIsvcWatch, runtimeControllerEvents); err != nil {
+	}).SetupWithManager(mgr, enableIsvcWatch, runtimeControllerEvents, enableCSRWatch); err != nil {
```
njhill, line 424: I think we probably want to add this as a ServingRuntimeReconciler field instead and also use it to control whether we Get/List the CSRs in various places (instead of using `HasNamespaceAccess` for that).

njhill (commented):
Thanks @chinhuang007, the updates look great... I've added a few more comments and resolved the prior comments that have now been addressed.

controllers/modelmesh/modelmesh.go
```diff
@@ -36,11 +36,14 @@ const ModelMeshEtcdPrefix = "mm"
...
-	Namespace          string
-	Owner              *kserveapi.ServingRuntime
+	ServiceName string
+	ServicePort uint16
+	Name        string
+	Namespace   string
+	// The Owner field can be nil in the CSR case
+	Owner              mf.Owner
+	SRSpec             *kserveapi.ServingRuntimeSpec
+	SRAnnotations      map[string]string
```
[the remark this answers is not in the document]
chinhuang007: Okay, will look into remove the separate annotations field.

controllers/servingruntime_controller.go
```diff
@@ -369,23 +414,47 @@ func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Co
 		return nil, err
 	}
 
+	// list all cluster serving runtimes
+	csrs := &kserveapi.ClusterServingRuntimeList{}
+	if r.HasNamespaceAccess {
+		if err := r.Client.List(ctx, csrs); err != nil {
+			return nil, err
+		}
+	}
```
njhill, line 423: nit: could move this below and include the `csrs` declaration and for loop within the if

main.go
```diff
@@ -76,7 +77,9 @@ const (
 	LeaderLockName                 = "modelmesh-controller-leader-lock"
 	LeaderForLifeLockName          = "modelmesh-controller-leader-for-life-lock"
 	EnableInferenceServiceEnvVar   = "ENABLE_ISVC_WATCH"
+	EnableClusterSrvngRntmEnvVar   = "ENABLE_CSR_WATCH"
```
njhill, line 80: Better to just use full words?
```suggestion
	EnableClusterServingRuntimeEnvVar   = "ENABLE_CSR_WATCH"
```

chinhuang007:
@njhill Most comments should be addressed, except the install/kustomization changes. Please take one more look. I am still investigating the FVT issues in Github Actions and think we might want to have a separate PR to install ClusterServingRuntimes by default, which will involve FVT changes as well.

controllers/servingruntime_controller.go
```diff
@@ -404,57 +406,55 @@ func runtimeSupportsPredictor(rt *kserveapi.ServingRuntimeSpec, p *api.Predictor
...
+		}
+		for i := range csrs.Items {
+			crt := &csrs.Items[i]
+			if crt.Spec.IsMultiModelRuntime() && runtimeSupportsPredictor(&crt.Spec, p, restProxyEnabled, crt.Name) {
+				if _, ok := srnns[crt.Name]; !ok {
+					srnns[crt.Name] = types.NamespacedName{
+						Name:      crt.Name,
+						Namespace: p.Namespace,
+					}
+				}
```
njhill, line 446: If you move the CSR part here before the regular SRs, this can just be a map addition, no need to have the existence check first.

controllers/servingruntime_controller.go
```diff
@@ -404,57 +406,55 @@ func runtimeSupportsPredictor(rt *kserveapi.ServingRuntimeSpec, p *api.Predictor
 	return true
 }
 
-// getRuntimesSupportingPredictor returns a list of keys for runtimes that support the predictor p
+// getRuntimesSupportingPredictor returns a map of keys for runtimes that support the predictor p
 //
 // A predictor may be supported by multiple runtimes.
-func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Context, p *api.Predictor) ([]types.NamespacedName, error) {
+func (r *ServingRuntimeReconciler) getRuntimesSupportingPredictor(ctx context.Context, p *api.Predictor) (map[string]types.NamespacedName, error) {
```
njhill, line 412: I was thinking here just populate/return the `map[string]struct{}` (just the set of string names), the `NamespacedName`s can be constructed in the calling function.

njhill (commented):
Thanks @chinhuang007 it's looking great. Just a couple more outstanding comments (I've resolved all the ones that have been addressed, see those that still aren't marked resolved).

controllers/servingruntime_controller.go
chinhuang007: I guess you would rather have more map additions/updates then checks for existence. Sure, I can make that change.

config/rbac/cluster-scope/role.yaml
```diff
@@ -175,3 +175,24 @@ rules:
...
+      - update
+      - watch
+  - apiGroups:
+      - serving.kserve.io
+    resources:
+      - clusterservingruntimes/status
+    verbs:
+      - get
+      - patch
+      - update
```
njhill, line 198: Best to remove this file too if we aren't doing the install updates in this PR

controllers/modelmesh/runtime.go
```diff
@@ -100,7 +100,8 @@ func calculateModelDirSize(rt *kserveapi.ServingRuntime) *resource.Quantity {
 
 //Adds the provided runtime to the deployment
 func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error {
-	rt := m.Owner
+	rta := m.Owner.GetAnnotations()
```
njhill, line 103: I don't think there's any need to introduced dedicated variable here

controllers/modelmesh/runtime.go
```diff
@@ -248,43 +249,44 @@ func (m *Deployment) addRuntimeToDeployment(deployment *appsv1.Deployment) error
 			{}, {}, {}, {}, // allocate larger array to avoid reallocation
 		}[:8]
 
-		if mlc, ok := rt.Annotations["maxLoadingConcurrency"]; ok {
-			builtInAdapterContainer.Env = append(builtInAdapterContainer.Env, corev1.EnvVar{
-				Name:  "LOADING_CONCURRENCY",
-				Value: mlc,
-			})
-		}
+		if rta != nil {
```
njhill, line 252: I think this `if` could be removed now, could revert to be closer to the original code (just `owner` instead of `rt`)

controllers/servingruntime_controller.go
```diff
@@ -94,21 +94,47 @@ type runtimeInfo struct {
...
+			srSpecMap[crt.GetName()] = crt.Spec
+		}
+	}
+	// rt spec would override crt spec by design
+	for i := range runtimes.Items {
+		rt := &runtimes.Items[i]
+		if rt.Spec.IsMultiModelRuntime() {
+			srSpecMap[rt.GetName()] = rt.Spec
+		}
+	}
```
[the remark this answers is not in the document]
chinhuang007: srSpecs[rt.GetName()] = &runtimes.Items[i].Spec won't work because rt is not defined. It could become `srSpecs[runtimes.Items[i].GetName()] = &runtimes.Items[i].Spec` Not sure it is better that what it is now. I will leave it unchanged.

scripts/install.sh
```diff
@@ -324,7 +324,7 @@ if [[ ! -z $user_ns_array ]]; then
       echo "Kube namespace does not exist: $user_ns. Will skip."
     else
       kubectl label namespace ${user_ns} modelmesh-enabled="true" --overwrite
-      kubectl apply -f runtimes.yaml -n ${user_ns}
+      #kubectl apply -f runtimes.yaml -n ${user_ns}
```
njhill, line 327: @chinhuang007 still some change to revert here

controllers/servingruntime_controller.go
```diff
@@ -475,8 +545,37 @@ func (r *ServingRuntimeReconciler) runtimeRequestsForPredictor(p *api.Predictor,
...
+	if err := r.Client.List(context.TODO(), list); err != nil || len(list.Items) == 0 {
+		return []reconcile.Request{}
 	}
+	requests := make([]reconcile.Request, 0, len(list.Items))
+
+	for i := range list.Items {
+		ns := &list.Items[i]
+		mme, err := modelMeshEnabled2(context.TODO(), ns.Name, r.ControllerNamespace, r.Client, r.HasNamespaceAccess)
+		if err == nil && mme {
+
```
njhill, line 572: Remove empty line?

controllers/servingruntime_controller.go
```diff
@@ -323,32 +373,33 @@ func (r *ServingRuntimeReconciler) determineReplicasAndRequeueDuration(ctx conte
...
+func (r *ServingRuntimeReconciler) runtimeHasPredictors(ctx context.Context, rt *kserveapi.ServingRuntimeSpec, rtName types.NamespacedName) (bool, error) {
 	restProxyEnabled := r.ConfigProvider.GetConfig().RESTProxy.Enabled
 	f := func(p *api.Predictor) bool {
-		return runtimeSupportsPredictor(rt, p, restProxyEnabled)
+		return runtimeSupportsPredictor(rt, p, restProxyEnabled, rtName.Name)
 	}
 
 	for _, pr := range r.RegistryMap {
-		if found, err := pr.Find(ctx, rt.GetNamespace(), f); found || err != nil {
+		//if found, err := pr.Find(ctx, rt.GetNamespace(), f); found || err != nil {
```
njhill, line 392: Remove commented line

controllers/servingruntime_controller.go
```diff
@@ -277,7 +326,8 @@ func (r *ServingRuntimeReconciler) determineReplicasAndRequeueDuration(ctx conte
 		r.runtimeInfoMap = make(map[types.NamespacedName]*runtimeInfo)
 	}
 
-	runtimeInfoMapKey := client.ObjectKeyFromObject(rt)
+	// runtimeInfoMapKey := client.ObjectKeyFromObject(rt)
```
njhill, line 329: Remove commented line

controllers/modelmesh/runtime_test.go
```diff
@@ -117,6 +118,7 @@ func TestOverlayMockRuntime(t *testing.T) {
 	}
 
 	m := newMockModelMeshDeployment(t, v)
+
```
njhill, line 121: Remove empty line

controllers/modelmesh/cluster_config.go
```diff
@@ -92,16 +94,18 @@ func (cc ClusterConfig) Reconcile(ctx context.Context, namespace string, cl clie
 }
 
 // Add constraint data to the provided config map
-func (cc ClusterConfig) addConstraints(rts *kserveapi.ServingRuntimeList, m *corev1.ConfigMap, restProxyEnabled bool) {
-	b := calculateConstraintData(rts.Items, restProxyEnabled)
+func (cc ClusterConfig) addConstraints(srSpecs map[string]*kserveapi.ServingRuntimeSpec, m *corev1.ConfigMap, restProxyEnabled bool) {
+	//b := calculateConstraintData(rts.Items, restProxyEnabled)
```
njhill, line 98: Remove commented line

controllers/modelmesh/cluster_config.go
```diff
@@ -54,18 +54,20 @@ var dataPlaneApiJsonConfigBytes = []byte(`{
...
-	Scheme   *runtime.Scheme
+	SRSpecs map[string]*kserveapi.ServingRuntimeSpec
+	Scheme  *runtime.Scheme
 }
 
 func (cc ClusterConfig) Reconcile(ctx context.Context, namespace string, cl client.Client, cfg *config.Config) error {
 	m := &corev1.ConfigMap{}
 	err := cl.Get(ctx, types.NamespacedName{Name: InternalConfigMapName, Namespace: namespace}, m)
 	notfound := errors.IsNotFound(err)
+
```
njhill, line 65: Remove empty line

njhill (commented):
Thanks @chinhuang007, just a little bit more cleanup...

njhill (approved):
Thanks @chinhuang007 for your hard work and patience!

PR #246, opened 2022-09-15
chore: Automatically set kube context in development container

ckadner:
#### Motivation

When using the containerized development environment `make develop` to run FVT tests, one needs to configure access to a Kubernetes or OpenShift cluster from inside the container. Which has to be done for every `make develop` session. This can be tricky when cloud provider specific CLI tools are needed to connect and authenticate to a cluster.

Currently there is a short paragraph in the FVT README about how to export a minified kubeconfig file and create that inside the container. It is tedious to repeat those steps for each `make develop` session and depending on OS, shell environment, editors and possible text encoding issue it is also error prone.

#### Modifications

This PR proposes to automatically create the kubeconfig file in a local and git-ignored directory inside the local project and automatically mount it to the develop container. All the user then has to do is connect and authenticate to the cluster in the shell that will be running `make develop`. 
 
#### Result

Kubernetes context is ready inside the development container.

```Shell
# shell environment, outside the develop container has access to K8s cluster
[modelmesh-serving_ckadner]$ kubectl get pods

NAME                                        READY   STATUS    RESTARTS   AGE
pod/etcd                                    1/1     Running   0          17m
pod/minio                                   1/1     Running   0          17m
pod/modelmesh-controller-387aef25be-ftyqu   1/1     Running   0          17m

[modelmesh-serving_ckadner]$ make develop

./scripts/build_devimage.sh
Pulling dev image kserve/modelmesh-controller-develop:6be58b09c25833c1...
Building dev image kserve/modelmesh-controller-develop:6be58b09c25833c1...
Image kserve/modelmesh-controller-develop:6be58b09c25833c1 has 14 layers
Tagging dev image kserve/modelmesh-controller-develop:6be58b09c25833c1 as latest
./scripts/develop.sh
[root@17c121286549 workspace]# kubectl get pods
NAME                                        READY   STATUS    RESTARTS   AGE
pod/etcd                                    1/1     Running   0          18m
pod/minio                                   1/1     Running   0          18m
pod/modelmesh-controller-387aef25be-ftyqu   1/1     Running   0          18m
[root@17c121286549 workspace]# 
```

/cc @njhill

### DIFF
--- a/scripts/develop.sh
+++ b/scripts/develop.sh
@@ -57,9 +57,14 @@ cd "${DIR}/.."
 # Make sure .bash_history exists and is a file
 touch .bash_history
 
+# create local copy of a kube-config file
+mkdir -p .kube/
+kubectl config view --minify --flatten 2> /dev/null > .kube/config
+
 declare -a docker_run_args=(
   -v "${PWD}:/workspace"
   -v "${PWD}/.bash_history:/root/.bash_history"
+  -v "${PWD}/.kube/config:/root/.kube/config"
   -v "/var/run/docker.sock:/var/run/docker.sock"
 )
 
--- a/scripts/develop.sh
+++ b/scripts/develop.sh
@@ -54,17 +54,19 @@ eval set -- "$PARAMS"
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 cd "${DIR}/.."
 
-# Make sure .bash_history exists and is a file
-touch .bash_history
+# store local development files in .dev directory
+mkdir -p .dev/
 
-# create local copy of a kube-config file
-mkdir -p .kube/
-kubectl config view --minify --flatten 2> /dev/null > .kube/config
+# Make sure .bash_history exists and is a file
+touch .dev/.bash_history
+
+# create a minified flattened local copy of the kube config
+kubectl config view --minify --flatten 2> /dev/null > .dev/.kube_config
 
 declare -a docker_run_args=(
   -v "${PWD}:/workspace"
-  -v "${PWD}/.bash_history:/root/.bash_history"
-  -v "${PWD}/.kube/config:/root/.kube/config"
+  -v "${PWD}/.dev/.bash_history:/root/.bash_history"
+  -v "${PWD}/.dev/.kube_config:/root/.kube/config"
   -v "/var/run/docker.sock:/var/run/docker.sock"
 )
 
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ bin
 devbuild
 .develop_image_name
 .bash_history
+.kube
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,4 @@ bin
 # Modelmesh development related artifacts
 devbuild
 .develop_image_name
-.bash_history
-.kube
+.dev/
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -22,12 +22,11 @@ kubectl create ns modelmesh-serving
 This installs the `modelmesh-controller` and dependencies in the `modelmesh-serving` namespace. The `minio` pod that this deploys
 contains special test images that are used in the functional tests.
 
-## Building development image
-
-A dockerized development environment is provided to help set up dependencies for testing, linting, and code generating. Using this environment is suggested as this is what the GitHub Actions workflows use. To create the development image, perform the following:
+If you already deployed ModelMesh Serving on a Kubernetes or OpenShift cluster before and are reconnecting to it now,
+make sure to set the default namespace to `modelmesh-serving`.
 
 ```shell
-make build.develop
+kubectl config set-context --current --namespace=modelmesh-serving
 ```
 
 ## Building and updating controller image
@@ -58,6 +57,26 @@ you will need to restart the controller pod. This can be done through the follow
 ```shell
 kubectl rollout restart deploy modelmesh-controller
 ```
+
+## Building the developer image
+
+A dockerized development environment is provided to help set up dependencies for testing, linting, and code generating.
+Using this environment is suggested as this is what the GitHub Actions workflows use.
+To create the development image, perform the following:
+
+```shell
+make build.develop
+```
+
+## Using the developer image for linting and testing
+
+To use the dockerized development environment run:
+
+```shell
+make develop
+```
+
+Then, from inside the developer container, proceed to run the linting, code generation, and testing as described below.
 
 ## Formatting and linting code
 
@@ -97,5 +116,5 @@ To run them, do the following:
 make fvt
 ```
 
-**Note**: sometimes the tests can fail on the first run because pulling the serving runtime images can take a while, causing a timeout.
-Just try again after the pulling is done.
+**Note**: sometimes the tests can fail on the first run because pulling the serving runtime images can take a while,
+causing a timeout. Just try again after the pulling is done.
--- a/fvt/README.md
+++ b/fvt/README.md
@@ -29,33 +29,47 @@ The FVTs rely on a set of models existing in a configured `localMinIO` storage. 
 
 If starting with a fresh namespace, install ModelMesh Serving configured for the FVTs with:
 
-```
+```Shell
 ./scripts/install.sh --namespace modelmesh-serving --fvt --dev-mode-logging
 ```
 
 To re-configure an existing quick-start instance for FVTs, run:
 
-```
+```Shell
 kubectl apply -f config/dependencies/fvt.yaml
 ```
 
 ### Development Environment
 
-The FVTs run using the `ginkgo` CLI tool and need `kubectl` configuration to the cluster. It is recommended to use the containerized development environment to run the FVTs. First build the environment with:
+The FVTs run using the `ginkgo` CLI tool and need `kubectl` configured to communicate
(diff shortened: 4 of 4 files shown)


scripts/develop.sh
```diff
@@ -57,9 +57,14 @@ cd "${DIR}/.."
...
 touch .bash_history
 
+# create local copy of a kube-config file
+mkdir -p .kube/
+kubectl config view --minify --flatten 2> /dev/null > .kube/config
+
 declare -a docker_run_args=(
   -v "${PWD}:/workspace"
   -v "${PWD}/.bash_history:/root/.bash_history"
+  -v "${PWD}/.kube/config:/root/.kube/config"
```
njhill, line 67: Thanks @ckadner this looks good in principle. But why does the kube context need to be saved separately? Why not just mount the user's `.kube` directory to the right place so that the existing context is used directly:
```suggestion
declare -a docker_run_args=(
  -v "${PWD}:/workspace"
  -v "${HOME}/.bash_history:/root/.bash_history"
  -v "${HOME}/.kube:/root/.kube"
```
(I think the `.bash_history` mount should be fixed too)

ckadner: Hi @njhill -- thanks for your review!

> Why not just mount the user's `.kube` directory to the right place so that the existing context is used directly

Because different cloud providers use various CLI tools to connect to K8s and OCP clusters, some of which do not create `~/.kube/config` files (i.e. IKS, OpenShift on IBM Cloud, OCP on Fyre).  In each of those cases, though, you can create that minified kubeconfig file and it works well.

> I think the .bash_history mount should be fixed too

My intention for the bashhistory was to keep only commands that are relevant to ModelMesh dev as opposed to all other user command histroy (mine spans a wide range of completely unrelated commands that I do not want to click through)

njhill: > Because different cloud providers use various CLI tools to connect to K8s and OCP clusters, some of which do not create ~/.kube/config files (i.e. IKS, OpenShift on IBM Cloud, OCP on Fyre). In each of those cases, though, you can create that minified kubeconfig file and it works well.

@ckadner I could very well be wrong but are you sure about this? The non-`kubectl` CLI tool you're referring to here is `oc` right? That shares the same kubectl context/config (see [here](https://docs.openshift.com/container-platform/4.9/cli_reference/openshift_cli/managing-cli-profiles.html#about-switches-between-cli-profiles_managing-cli-profiles)). This is only a client-side thing, it shouldn't matter what kind of Kubernetes cluster you're connecting to...

> My intention for the .bashhistory was to keep only those commands that are relevant to ModelMesh development.

Ah ok now I understand the `${PWD}` rationale for that, sounds reasonable, thanks!

ckadner: Hi @njhill -- yes `oc` does not use/create a `~/.kube/config` file (see below) and for IKS I am using the IKS CLI and a tool to switch between clusters by pointing to different `kube-config` files which the `ibmcloud` CLI stores under `${HOME}/.bluemix/plugins/container-service/clusters/...` i.e.
```Shell
KUBECONFIG=/Users/ckadner/.bluemix/plugins/container-service/clusters/ckadner-modelmesh-dev-cbu3tugd0r5j02te049g/kube-config-aaa00-ckadner-modelmesh-dev.yml
```
This is very useful when I am working on different clusters in different terminal shells at the same time, where I need different kube contexts for each of the terminals.

For OpenShift on IBM Cloud or OCP on Fyre, I see this behavior:

```Bash
[modelmesh-serving_ckadner] (main *=)$ oc login --token=sha256~u_o6N****zznde_3rw --server=https://c104-e.us-south.containers.cloud.ibm.com:30276

Logged into "https://c104-e.us-south.containers.cloud.ibm.com:30276" as "IAM#ckadner@us.ibm.com" using the token provided.

You have access to 67 projects, the list has been suppressed. You can list all projects with 'oc projects'

Using project "default".


[modelmesh-serving]$ kubectl get pods -n kube-system | head

NAME                                             READY   STATUS    RESTARTS   AGE
ibm-file-plugin-684495896f-x6vsp                 1/1     Running   0          14d
ibm-keepalived-watcher-6zg84                     1/1     Running   0          14d
ibm-keepalived-watcher-9zqw9                     1/1     Running   0          14d
ibm-master-proxy-static-10.87.76.105             2/2     Running   2          28d
ibm-master-proxy-static-10.87.76.117             2/2     Running   2          28d
ibm-storage-metrics-agent-5c89ffc69f-hqd4h       1/1     Running   0          4d20h
ibm-storage-watcher-77847bfb4c-xp5jn             1/1     Running   0          14d
ibmcloud-block-storage-driver-n99lg              1/1     Running   0          14d
ibmcloud-block-storage-driver-rjtrg              1/1     Running   0          14d


[modelmesh-serving]$ cat ~/.kube/config

cat: /Users/ckadner/.kube/config: No such file or directory


[modelmesh-serving]$ cat scripts/develop.sh | grep kube

  -v "${HOME}/.kube:/root/.kube"


[modelmesh-serving]$ make develop

./scripts/build_devimage.sh
Pulling dev image kserve/modelmesh-controller-develop:6be58b09c25833c1...
Building dev image kserve/modelmesh-controller-develop:6be58b09c25833c1
[+] Building 1.0s (10/10) FINISHED                                                                                                                                       
...
Image kserve/modelmesh-controller-develop:6be58b09c25833c1 has 14 layers
Tagging dev image kserve/modelmesh-controller-develop:6be58b09c25833c1 as latest
./scripts/develop.sh
[root@72d9bf620910 workspace]# kubectl get pods
error: Missing or incomplete configuration info.  Please point to an existing, complete config file:


  1. Via the command-line flag --kubeconfig
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config

To view or setup config directly use the 'config' command.
[root@72d9bf620910 workspace]# 
```

njhill:
@ckadner how about
```
-v "${KUBECONFIG:-${PWD}/.kube/config}:/root/.kube/config"
```

ckadner:
> @ckadner how about
> 
> ```
> -v "${KUBECONFIG:-${PWD}/.kube/config}:/root/.kube/config"
> ```

Hi @njhill -- I think you meant the user's `~/.kube/config` like this?

```
  -v "${KUBECONFIG:-${HOME}/.kube/config}:/root/.kube/config"
```

Unfortunately also does not work since, for IKS at least, the kubeconfig files under `${HOME}/.bluemix/plugins/container-service/clusters/` have local absolute file references, which will not be present if only mounting the the one kubeconfig file without the whole `${HOME}/.bluemix/` folder also mounted:

```
[root@5f681a6f418d workspace]# kubectl get pods
error: unable to read certificate-authority /Users/ckadner/.bluemix/plugins/container-service/clusters/ckadner-modelmesh-dev-cbu3tugd0r5j02te049g/ca-aaa00-ckadner-modelmesh-dev.pem for ckadner-modelmesh-dev/cbu3tugd0r5j02te049g due to open /Users/ckadner/.bluemix/plugins/container-service/clusters/ckadner-modelmesh-dev-cbu3tugd0r5j02te049g/ca-aaa00-ckadner-modelmesh-dev.pem: no such file or directory
```

Does the minified kubeconfig approach in this PR not work for you? Or you don't like creating another local file? We could put that into the user's `$TMP` dir instead or tuck away all local developer related artifacts into a `.dev` folder, like `.bash_history` and `develop_image_name` ...

njhill:
> I think you meant the user's ~/.kube/config like this?

Yes, sorry, that's what I meant!

> the kubeconfig files under `${HOME}/.bluemix/plugins/container-service/clusters/` have local absolute file references,

Ah ok, fair enough

> Does the minified kubeconfig approach in this PR not work for you?

No it's fine, it just seemed cleaner to me to link to the outside config if possible. But I understand the complications now so this seems fine. I guess my only concern is that this would overwrite an existing `.kube/config` file if it was run in the user's home dir.

ckadner:
> I guess my only concern is that this would overwrite an existing .kube/config file if it was run in the user's home dir.

Right, but it should not break anything since it would rewrite it with the same kubeconfig file contents -- just "minified" :-)

We could instead save all local development artifacts in a `.dev` folder inside the users current folder, which we assume to be the local clone of modelmesh-serving. Artifacts like `.bash_history` and `.develop_image_name` and the now new `.kube/config` file. WDYT?

ckadner:
> > Right, but it should not break anything since it would rewrite it with the same kubeconfig file contents -- just "minified" :-)
> 
> What if the saved config comes from a different location (due to env var setting)?
> 
> > We could instead save all local development artifacts in a .dev folder inside the users current folder, which we assume to be the local clone of modelmesh-serving. Artifacts like `.bash_history` and `.develop_image_name` and the now new `.kube/config` file.
> 
> That sounds like a good idea to me!

@njhill -- I moved the `.bash_history` and `.kube_config` file inside a `.dev` folder:

- [Move .bash_history and .kube/config files into .dev directory](https://github.com/kserve/modelmesh-serving/pull/246/commits/3fbdf19) [[3fbdf19](https://github.com/kserve/modelmesh-serving/pull/246/commits/3fbdf19)]

And added a little bit of text to the developer doc:

- [Update developer.md](https://github.com/kserve/modelmesh-serving/pull/246/commits/7677434) [[7677434](https://github.com/kserve/modelmesh-serving/pull/246/commits/7677434)]

Could you you do one more review of this PR? Thank you!

njhill (approved):
Thanks @ckadner!

PR #252, opened 2022-09-29
feat: Update install to support ClusterServingRuntime

chinhuang007:
Update install and config to support ClusterServingRuntime
- update install script to setup SRs or CSRs
- add rbac to access CSRs
- add and update install yamls
- update kserve ServingRuntimes and ClusterServingRuntimes yaml
- update fvt to work with either cluster or namespace scope mode
- update unit tests to work with updated yamls

We will need a separate PR to cover more FVT scenarios for the cluster scope mode with ClusterServingRuntimes

### DIFF
--- a/config/cluster-runtimes/kustomization.yaml
+++ b/config/cluster-runtimes/kustomization.yaml
@@ -1,0 +1,14 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../runtimes
+patchesJson6902:
+  - target:
+      group: serving.kserve.io
+      version: v1alpha1
+      kind: ServingRuntime
+      name: ".*"
+    patch: |-
+      - op: replace
+        path: /kind
+        value: ClusterServingRuntime
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -18,8 +18,7 @@ resources:
   # Including creation of Service here for now, will later be done automatically
   - bases/serving.kserve.io_predictors.yaml
   - bases/serving.kserve.io_inferenceservices.yaml
-  # ClusterServingRuntime not yet supported by modelmesh-serving
-  #  - bases/serving.kserve.io_clusterservingruntimes.yaml
+  - bases/serving.kserve.io_clusterservingruntimes.yaml
   - bases/serving.kserve.io_servingruntimes.yaml
   # +kubebuilder:scaffold:crdkustomizeresource
 
--- a/controllers/servingruntime_controller_test.go
+++ b/controllers/servingruntime_controller_test.go
@@ -53,9 +53,9 @@ func waitForAndGetRuntimeDeployment(runtimeName string) *appsv1.Deployment {
 
 var _ = Describe("Sample Runtime", func() {
 	samplesToTest := []string{
-		"controllers/testdata/mlserver-0.x.yaml",
-		"controllers/testdata/triton-2.x.yaml",
-		"controllers/testdata/ovms-1.x.yaml",
+		"config/runtimes/mlserver-0.x.yaml",
+		"config/runtimes/triton-2.x.yaml",
+		"config/runtimes/ovms-1.x.yaml",
 	}
 	for _, f := range samplesToTest {
 		// capture the value in new variable for each iteration
@@ -69,7 +69,7 @@ var _ = Describe("Sample Runtime", func() {
 				m, err = mf.ManifestFrom(mf.Path(filepath.Join("..", sampleFilename)))
 				m.Client = mfc.NewClient(k8sClient)
 				Expect(err).ToNot(HaveOccurred())
-				m, err = m.Transform(mf.InjectNamespace(namespace))
+				m, err = m.Transform(convertToServingRuntime, mf.InjectNamespace(namespace))
 				Expect(err).ToNot(HaveOccurred())
 				err = m.Apply()
 				Expect(err).ToNot(HaveOccurred())
@@ -97,10 +97,10 @@ var _ = Describe("Prometheus metrics configuration", func() {
 		reconcilerConfig.Metrics.Enabled = true
 
 		By("create a sample runtime")
-		m, err = mf.ManifestFrom(mf.Path("../controllers/testdata/mlserver-0.x.yaml"))
+		m, err = mf.ManifestFrom(mf.Path("../config/runtimes/mlserver-0.x.yaml"))
 		m.Client = mfc.NewClient(k8sClient)
 		Expect(err).ToNot(HaveOccurred())
-		m, err = m.Transform(mf.InjectNamespace(namespace))
+		m, err = m.Transform(convertToServingRuntime, mf.InjectNamespace(namespace))
 		Expect(err).ToNot(HaveOccurred())
 		err = m.Apply()
 		Expect(err).ToNot(HaveOccurred())
@@ -176,10 +176,10 @@ var _ = Describe("REST Proxy configuration", func() {
 		reconcilerConfig.RESTProxy.Enabled = true
 
 		By("create a sample runtime")
-		m, err = mf.ManifestFrom(mf.Path("../controllers/testdata/mlserver-0.x.yaml"))
+		m, err = mf.ManifestFrom(mf.Path("../config/runtimes/mlserver-0.x.yaml"))
 		m.Client = mfc.NewClient(k8sClient)
 		Expect(err).ToNot(HaveOccurred())
-		m, err = m.Transform(mf.InjectNamespace(namespace))
+		m, err = m.Transform(convertToServingRuntime, mf.InjectNamespace(namespace))
 		Expect(err).ToNot(HaveOccurred())
 		err = m.Apply()
 		Expect(err).ToNot(HaveOccurred())
@@ -193,3 +193,10 @@ var _ = Describe("REST Proxy configuration", func() {
 		Expect(d).To(SnapshotMatcher())
 	})
 })
+
+func convertToServingRuntime(resource *unstructured.Unstructured) error {
+	if resource.GetKind() == "ClusterServingRuntime" {
+		resource.SetKind("ServingRuntime")
+	}
+	return nil
+}
--- a/controllers/servingruntime_controller_test.go
+++ b/controllers/servingruntime_controller_test.go
@@ -53,9 +53,9 @@ func waitForAndGetRuntimeDeployment(runtimeName string) *appsv1.Deployment {
 
 var _ = Describe("Sample Runtime", func() {
 	samplesToTest := []string{
-		"config/runtimes/mlserver-0.x.yaml",
-		"config/runtimes/triton-2.x.yaml",
-		"config/runtimes/ovms-1.x.yaml",
+		"controllers/testdata/mlserver-0.x.yaml",
+		"controllers/testdata/triton-2.x.yaml",
+		"controllers/testdata/ovms-1.x.yaml",
 	}
 	for _, f := range samplesToTest {
 		// capture the value in new variable for each iteration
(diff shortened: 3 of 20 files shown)


fvt/predictor/predictor_suite_test.go
```diff
@@ -60,10 +62,14 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-	// confirm 3 serving runtimes exist
+	// confirm 3 serving runtimes exist in namespace scope mode
 	list, err := FVTClientInstance.ListServingRuntimes(metav1.ListOptions{})
 	Expect(err).ToNot(HaveOccurred())
-	Expect(list.Items).To(HaveLen(3))
+	if NameSpaceScopeMode {
+		Expect(list.Items).To(HaveLen(3))
+	} else {
+		Expect(list.Items).To(HaveLen(0))
+	}
```
njhill, line 72: For these, wouldn't it be better to list either SRs or CSRs depending on namespace/cluster mode? I.e. the list should still contain 3 in either case.

config/cluster-runtimes/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+patchesJson6902:
+  - target:
+      group: serving.kserve.io
+      version: v1alpha1
+      kind: ServingRuntime
+      name: ".*"
+    patch: |-
+      - op: replace
+        path: /kind
+        value: ClusterServingRuntime
```
njhill, line 14: If cluster scope is the default install mode, I'm wondering whether we should invert this - change the SRs in `runtimes` to be CSRs, and have this overlay change them to SRs for the namespace scope install case?

scripts/deploy/iks/test-fvt.sh
```diff
@@ -49,19 +48,22 @@ run_fvt() {
...
 
   if [[ $(grep "Test Suite Passed" fvt.out) ]]; then
-    export NAMESPACE="modelmesh-user"
-    ginkgo -v --progress --fail-fast -p fvt/predictor fvt/scaleToZero --timeout 40m > fvt.out
-    cat fvt.out
-    if [[ $(grep "Test Suite Passed" fvt.out) ]]; then
       REV=0
-    fi
   fi
-
```
njhill, line 64: I don't completely understand the changes in this file, are they temporary?

config/crd/kustomization.yaml
```diff
@@ -18,8 +18,7 @@ resources:
   # Including creation of Service here for now, will later be done automatically
   - bases/serving.kserve.io_predictors.yaml
   - bases/serving.kserve.io_inferenceservices.yaml
-  # ClusterServingRuntime not yet supported by modelmesh-serving
-  #  - bases/serving.kserve.io_clusterservingruntimes.yaml
+  - bases/serving.kserve.io_clusterservingruntimes.yaml
```
njhill, line 21: I'm wondering whether we should include this in namespace scope case.

njhill (commented):
Thanks @chinhuang007 this looks pretty good, just a few comments.

fvt/predictor/predictor_suite_test.go
chinhuang007: As mentioned in commit comments, I try to minimize the FVT changes to  test CSRs, for which I suggest to create a new PR.

config/cluster-runtimes/kustomization.yaml
chinhuang007: Again, due to existing FVTs and unit test depend on these to be SR, I try to minimize the changes. A separate PR should address your concern.

scripts/deploy/iks/test-fvt.sh
chinhuang007: Yes, this is temporarily to make the test only for the controller namespace in the namespace scope mode, until we have the FVTs all figured out for CSRs in cluster scope mode.

config/crd/kustomization.yaml
chinhuang007: Good question. Since we won't do anything with CSRs in namespace scope mode, probably can exclude it during install.

controllers/servingruntime_controller.go
```diff
@@ -138,7 +138,7 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	// Delete etcd secret when there is no ServingRuntimes in a namespace
 	etcdSecretName := cfg.GetEtcdSecretName()
-	if len(runtimes.Items) == 0 {
+	if len(srSpecs) == 0 {
```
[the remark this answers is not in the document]
njhill: @chinhuang007 this is really a fix to the prior PR, we could just open a separate PR with just this and get it merged first.

chinhuang007:
@njhill I made some changes to address comments. Please take one more look.

config/crd/kustomization.yaml
njhill: @chinhuang007 WDYT about excluding this for namespace mode? We could just have a separate cluster_crd kustomization which includes this one + the CSR CRD

controllers/servingruntime_controller_test.go
```diff
@@ -53,9 +53,9 @@ func waitForAndGetRuntimeDeployment(runtimeName string) *appsv1.Deployment {
 
 var _ = Describe("Sample Runtime", func() {
 	samplesToTest := []string{
-		"config/runtimes/mlserver-0.x.yaml",
-		"config/runtimes/triton-2.x.yaml",
-		"config/runtimes/ovms-1.x.yaml",
+		"controllers/testdata/mlserver-0.x.yaml",
+		"controllers/testdata/triton-2.x.yaml",
+		"controllers/testdata/ovms-1.x.yaml",
```
njhill, line 58: I think it would be a lot cleaner for these to convert it from CSR to SR programmatically. Then we won't need to copy files around.

And it should be easy with manifestival, you can just have a top level function like
```go
func convertToServingRuntime(resource *unstructured.Unstructured) error {
	if resource.GetKind() == "ClusterServingRuntime" {
		resource.SetKind("ServingRuntime")
	}
	return nil
}
```
and then just add it to the existing Transform() call where each of the CRs are loaded, i.e.:
```go
		By("create the runtime")
		m, err = mf.ManifestFrom(mf.Path(filepath.Join("..", sampleFilename)))
		m.Client = mfc.NewClient(k8sClient)
		Expect(err).ToNot(HaveOccurred())
		m, err = m.Transform(convertToServingRuntime, mf.InjectNamespace(namespace))
		Expect(err).ToNot(HaveOccurred())
		// ...
```

fvt/predictor/predictor_suite_test.go
```diff
@@ -60,8 +62,12 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	createFVTClient()
 
-	// confirm 3 serving runtimes exist
-	list, err := FVTClientInstance.ListServingRuntimes(metav1.ListOptions{})
+	// confirm 3 cluster serving runtimes or serving runtimes exist
+	f := FVTClientInstance.ListClusterServingRuntimes
+	if NameSpaceScopeMode {
+		f = FVTClientInstance.ListServingRuntimes
+	}
+	list, err := f(metav1.ListOptions{})
```
njhill, line 70: Maybe this is a bit clearer?:
```suggestion
	var err error
	var list *unstructured.UnstructuredList
	if NameSpaceScopeMode {
		list, err = FVTClientInstance.ListServingRuntimes(metav1.ListOptions{})
	} else {
		list, err = FVTClientInstance.ListClusterServingRuntimes(metav1.ListOptions{})
	}
```

fvt/scaleToZero/scaleToZero_suite_test.go
```diff
@@ -51,16 +51,22 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	Log.Info("FVTClientInstance created", "client", FVTClientInstance)
 
-	// confirm 3 serving runtimes exist
-	list, err := FVTClientInstance.ListServingRuntimes(metav1.ListOptions{})
+	// confirm 3 cluster serving runtimes or serving runtimes exist
+	f := FVTClientInstance.ListClusterServingRuntimes
+	if NameSpaceScopeMode {
+		f = FVTClientInstance.ListServingRuntimes
+	}
+	list, err := f(metav1.ListOptions{})
```
njhill, line 69: Same comment as above, except `err` is already declared here

scripts/deploy/iks/test-fvt.sh
```diff
@@ -62,6 +61,9 @@ run_fvt() {
     fi
   fi
 
+  if [[ $(grep "Test Suite Passed" fvt.out) ]]; then
+      REV=0
+  fi
```
njhill, line 66: I don't understand why this is added here?

scripts/install.sh
```diff
@@ -239,6 +239,7 @@ if [[ $delete == "true" ]]; then
   info "Deleting any previous ModelMesh Serving instances and older CRD with serving.kserve.io api group name"
   kubectl delete crd/predictors.serving.kserve.io --ignore-not-found=true
   kubectl delete crd/servingruntimes.serving.kserve.io --ignore-not-found=true
+  kubectl delete crd/clusterservingruntimes.serving.kserve.io --ignore-not-found=true
```
njhill, line 242: Should move this into the `namespace_scope != true` block below I think?

scripts/install.sh
```diff
@@ -296,6 +298,7 @@ fi
 if [[ $namespace_scope_mode == "true" ]]; then
   info "Enabling namespace scope mode"
   kubectl set env deploy/modelmesh-controller NAMESPACE_SCOPE=true
+  sed -i 's/#- bases\/serving.kserve.io_clusterservingruntimes.yaml/- bases\/serving.kserve.io_clusterservingruntimes.yaml/g' crd/kustomization.yaml
```
njhill, line 301: I don't think this is needed here, it's already done above?

scripts/install.sh
```diff
@@ -281,6 +282,7 @@ kubectl get secret storage-config
 info "Installing ModelMesh Serving RBACs (namespace_scope_mode=$namespace_scope_mode)"
 if [[ $namespace_scope_mode == "true" ]]; then
   kustomize build rbac/namespace-scope | kubectl apply -f -
+  sed -i 's/- bases\/serving.kserve.io_clusterservingruntimes.yaml/#- bases\/serving.kserve.io_clusterservingruntimes.yaml/g' crd/kustomization.yaml
```
njhill, line 285: Ah I see, you didn't make a separate kustomization for this. I guess this is ok. Maybe add a comment above this line saying that we won't install the CSR CRD in namespace mode?

scripts/install.sh
```diff
@@ -312,10 +315,13 @@ elif [[ -n "$kustomize_version" && "$kustomize_version" < "4.0.1" ]]; then
...
-kustomize build runtimes ${kustomize_load_restrictor_arg} | kubectl apply -f -
+if [[ $namespace_scope_mode == "true" ]]; then
+    kustomize build namespace-runtimes ${kustomize_load_restrictor_arg} | kubectl apply -f -
+else
+    kustomize build runtimes ${kustomize_load_restrictor_arg} | kubectl apply -f -
+fi
 
-if [[ ! -z $user_ns_array ]]; then
-  kustomize build runtimes ${kustomize_load_restrictor_arg} > runtimes.yaml
+if [[ $namespace_scope_mode == "false" ]] && [[ ! -z $user_ns_array ]]; then
```
njhill, line 324: maybe safer to do `!= "true"` instead of `== "false"` here?

njhill (commented):
Thanks @chinhuang007 it's looking pretty good, just a few more comments.

scripts/install.sh
chinhuang007: This is to reset crd/kustomization.yaml back to original in case the user wants to delete the namespaced installation and deploy a cluster scoped installation using the same set of files.

chinhuang007:
@njhill Thanks for the review. One more commit is pushed. Please take a look.

scripts/install.sh
njhill: Ah ok sorry I didn't notice that this was reverting.

controllers/servingruntime_controller_test.go
njhill: OK, it just feels like this is a bit more hacky/brittle.

njhill (approved):
Thanks @chinhuang007, LGTM. I pushed a couple more small changes to avoid another review roundtrip :)

PR #257, opened 2022-10-05
fix: Fix a couple of bugs for ClusterServingRuntime

chinhuang007:
Fix a couple of bugs for ClusterServingRuntime and also rename two variables to reflect the cluster scope mode is enabled

#### Motivation
Fix bugs and make code more readable

### DIFF
--- a/controllers/config/manifest.go
+++ b/controllers/config/manifest.go
@@ -54,7 +54,7 @@ func Apply(cl client.Client, owner metav1.Object, templatePath string, context i
 	return nil
 }
 
-func Delete(cl client.Client, owner metav1.Object, templatePath string, context interface{}, fns ...mf.Transformer) error {
+func Delete(cl client.Client, owner metav1.Object, templatePath string, context interface{}, namespace string, fns ...mf.Transformer) error {
 	m, err := mf.ManifestFrom(PathTemplateSource(templatePath, context))
 	if err != nil {
 		return err
@@ -64,9 +64,12 @@ func Delete(cl client.Client, owner metav1.Object, templatePath string, context 
 	if owner != nil {
 		asMfOwner := owner.(mf.Owner)
 		fns = append(fns, mf.InjectOwner(asMfOwner))
-		fns = append(fns, mf.InjectNamespace(asMfOwner.GetNamespace()))
+		if asMfOwner.GetNamespace() != "" {
+			fns = append(fns, mf.InjectNamespace(asMfOwner.GetNamespace()))
+		} else {
+			fns = append(fns, mf.InjectNamespace(namespace))
+		}
 	}
-
 	m, err = m.Transform(fns...)
 	if err != nil {
 		return err
--- a/controllers/config/manifest.go
+++ b/controllers/config/manifest.go
@@ -64,11 +64,7 @@ func Delete(cl client.Client, owner metav1.Object, templatePath string, context 
 	if owner != nil {
 		asMfOwner := owner.(mf.Owner)
 		fns = append(fns, mf.InjectOwner(asMfOwner))
-		if asMfOwner.GetNamespace() != "" {
-			fns = append(fns, mf.InjectNamespace(asMfOwner.GetNamespace()))
-		} else {
-			fns = append(fns, mf.InjectNamespace(namespace))
-		}
+		fns = append(fns, mf.InjectNamespace(namespace))
 	}
 	m, err = m.Transform(fns...)
 	if err != nil {
--- a/controllers/servingruntime_controller.go
+++ b/controllers/servingruntime_controller.go
@@ -71,8 +71,8 @@ type ServingRuntimeReconciler struct {
 	ConfigMapName       types.NamespacedName
 	ControllerName      string
 	ControllerNamespace string
-	// whether the controller has RBAC permission to read namespaces
-	HasNamespaceAccess bool
+	// whether the controller has cluster scope permissions
+	ClusterScope bool
 	// whether the controller is enabled to read and watch ClusterServingRuntimes
 	EnableCSRWatch bool
 	// store some information about current runtimes for making scaling decisions
@@ -98,7 +98,7 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	log := r.Log.WithValues("servingruntime", req.NamespacedName)
 	log.V(1).Info("ServingRuntime reconciler called")
 	// Make sure the namespace has serving enabled
-	mmEnabled, err := modelMeshEnabled2(ctx, req.Namespace, r.ControllerNamespace, r.Client, r.HasNamespaceAccess)
+	mmEnabled, err := modelMeshEnabled2(ctx, req.Namespace, r.ControllerNamespace, r.Client, r.ClusterScope)
 	if err != nil {
 		return RequeueResult, err
 	}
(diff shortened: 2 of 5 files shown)


controllers/servingruntime_controller.go
```diff
@@ -509,25 +509,55 @@ func (r *ServingRuntimeReconciler) SetupWithManager(mgr ctrl.Manager,
...
-		if filter == nil || filter(rt) {
-			requests = append(requests, reconcile.Request{
-				NamespacedName: types.NamespacedName{Name: rt.Name, Namespace: rt.Namespace},
-			})
+
+	srnns := make(map[types.NamespacedName]struct{})
+
+	if r.EnableCSRWatch {
+		csrs := &kserveapi.ClusterServingRuntimeList{}
+		if err := r.Client.List(context.TODO(), csrs); err != nil {
```
njhill, line 527: Log the error here (similar to above). I noticed a similar log line is also missing below in the `clusterServingRuntimeRequests` function (if there's an error getting namespaces).

controllers/servingruntime_controller.go
```diff
@@ -509,25 +509,55 @@ func (r *ServingRuntimeReconciler) SetupWithManager(mgr ctrl.Manager,
...
+			return []reconcile.Request{}
+		}
+		for i := range csrs.Items {
+			csr := &csrs.Items[i]
+			if csr.Spec.IsMultiModelRuntime() {
+				list := &corev1.NamespaceList{}
+				if err := r.Client.List(context.TODO(), list); err != nil {
+					r.Log.Error(err, "Error listing ClusterServingRuntimes to reconcile")
```
njhill, line 535: Incorrect error message here

controllers/servingruntime_controller.go
```diff
@@ -509,25 +509,55 @@ func (r *ServingRuntimeReconciler) SetupWithManager(mgr ctrl.Manager,
...
+	if r.EnableCSRWatch {
+		csrs := &kserveapi.ClusterServingRuntimeList{}
+		if err := r.Client.List(context.TODO(), csrs); err != nil {
+			return []reconcile.Request{}
+		}
+		for i := range csrs.Items {
+			csr := &csrs.Items[i]
+			if csr.Spec.IsMultiModelRuntime() {
+				list := &corev1.NamespaceList{}
+				if err := r.Client.List(context.TODO(), list); err != nil {
```
njhill, line 534: We should just retrieve the namespaces once before the loop, and only if the `namespace` parameter is empty. Something like:
```go
var namespaces []string
if namespace != "" {
    namespaces = []string{namespace}
} else {
   list := &corev1.NamespaceList{}
   if err := r.Client.List(context.TODO(), list); err != nil {
       // ...
   }
   for j := range list.Items {
        ns := &list.Items[j]
        if filter == nil || filter(ns) {
            namespaces = append(namespaces, ns)
        }
    }
}
```

then just use `namespaces` with the CSR loop.

controllers/servingruntime_controller.go
```diff
@@ -509,25 +509,55 @@ func (r *ServingRuntimeReconciler) SetupWithManager(mgr ctrl.Manager,
...
 	}
-	requests := make([]reconcile.Request, 0, len(list.Items))
-	for i := range list.Items {
-		rt := &list.Items[i]
-		if filter == nil || filter(rt) {
-			requests = append(requests, reconcile.Request{
-				NamespacedName: types.NamespacedName{Name: rt.Name, Namespace: rt.Namespace},
-			})
+
+	srnns := make(map[types.NamespacedName]struct{})
```
njhill, line 523: There's no need for this in the `!r.EnableCSRWatch` case so maybe move it within the `if` block and just handle the SRs like they are now if CSRs are disabled? i.e. different loop over SR list in both if and else cases.

njhill (commented):
Thanks @chinhuang007, just a couple of comments

controllers/servingruntime_controller.go
njhill: Also no need to call `modelMeshEnabled2`, it's done by the filter.

controllers/servingruntime_controller.go
```diff
@@ -509,25 +509,69 @@ func (r *ServingRuntimeReconciler) SetupWithManager(mgr ctrl.Manager,
...
-	requests := make([]reconcile.Request, 0, len(list.Items))
-	for i := range list.Items {
-		rt := &list.Items[i]
-		if filter == nil || filter(rt) {
-			requests = append(requests, reconcile.Request{
-				NamespacedName: types.NamespacedName{Name: rt.Name, Namespace: rt.Namespace},
-			})
+
+	var requests []reconcile.Request
+	if r.EnableCSRWatch {
```
njhill, line 524: I just realized it would be better here to also go to the `else` case if there are no CSRs.
i.e. something like
```go
var csrs *kserveapi.ClusterServingRuntimeList
if r.EnableCSRWatch {
    csrs = &kserveapi.ClusterServingRuntimeList{}
    if err := r.Client.List(context.TODO(), csrs); err != nil {
    	r.Log.Error(err, "Error listing ClusterServingRuntimes to reconcile")
	return []reconcile.Request{}
    } 
}
if csrs != nil && len(csrs.Items) > 0 {
    srnns := make(map[types.NamespacedName]struct{})
    // ...
```

njhill (commented):
Thanks @chinhuang007 looks great, just one more small thing I just thought of.

njhill (commented):
Thanks @chinhuang007, LGTM now, I guess we just need to track down how this nil deref is happening...

controllers/config/manifest.go
```diff
@@ -64,9 +64,12 @@ func Delete(cl client.Client, owner metav1.Object, templatePath string, context
 	if owner != nil {
 		asMfOwner := owner.(mf.Owner)
 		fns = append(fns, mf.InjectOwner(asMfOwner))
-		fns = append(fns, mf.InjectNamespace(asMfOwner.GetNamespace()))
+		if asMfOwner.GetNamespace() != "" {
+			fns = append(fns, mf.InjectNamespace(asMfOwner.GetNamespace()))
+		} else {
+			fns = append(fns, mf.InjectNamespace(namespace))
+		}
```
njhill, line 71:
```suggestion
		fns = append(fns, mf.InjectNamespace(namespace))
```
@chinhuang007 no need for the if/else here

controllers/servingruntime_controller.go
```diff
@@ -259,7 +259,7 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		AnnotationsMap:      cfg.RuntimePodAnnotations,
 		LabelsMap:           cfg.RuntimePodLabels,
 	}
-
+	log.V(1).Info("track deployment owner", "mmDeployment.Owner", mmDeployment.Owner)
```
njhill, line 262: Maybe remove this now?

njhill (commented):
Thanks @chinhuang007, latest fix looks good, but can just always set the provided namespace.

njhill:
@chinhuang007 incidentally I also noticed that setting a cluster-scoped owner here does nothing: https://github.com/manifestival/manifestival/blob/master/transform.go#L97-L100

which is unfortunate. I wonder whether that statement about it being broken is still true (it's pretty old). The [Kube docs](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents) don't mention anything about it.

njhill (approved):
Thanks @chinhuang007, LGTM!

chinhuang007:
It seems a bit odd that doc says " Namespaced dependents can specify cluster-scoped or namespaced owners." but the manifestival transform code does nothing to cluster-scoped owners.

njhill:
FWIW I opened an issue in the manifestival repo to ask about this: https://github.com/manifestival/manifestival/issues/90

PR #258, opened 2022-10-05
chore: Update to use latest OVMS 2022.2 image

njhill:
#### Motivation

Intel have released a new version of OpenVINO Model Server which now supports the KServe V2 prediction API (in addition to existing V1 API support): https://github.com/openvinotoolkit/model_server/releases/tag/v2022.2

#### Modifications

Update OVMS image tag from 2022.1 to 2022.2 in the ServingRuntime configuration.

#### Result

KServe V2 API can be used with OpenVINO Model Server and model-mesh.


chinhuang007 (approved):
Looks good!

PR #266, opened 2022-10-22
feat: Update makefile to support building with podman

anishasthana:
This PR will update the makefile and build scripts to allow developers to use podman as the build engine.
The default behaviour for the makefile will remain the same.

I ran the following to build images:

```
make build.develop ENGINE=podman
make build ENGINE=podman
```

I don't have a working docker env to test in.
cc @njhill

### DIFF
--- a/scripts/build_devimage.sh
+++ b/scripts/build_devimage.sh
@@ -19,6 +19,7 @@ set -euo pipefail
 CONTEXT_DIR=devbuild
 IMAGE_NAME=kserve/modelmesh-controller-develop
 DEV_DEPS="$0 Dockerfile.develop go.mod go.sum .pre-commit-config.yaml"
+ENGINE=${1:-docker}
 
 # command is shasum on osx
 SHASUM=sha1sum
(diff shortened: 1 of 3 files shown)


scripts/build_devimage.sh
```diff
@@ -19,6 +19,7 @@ set -euo pipefail
 CONTEXT_DIR=devbuild
 IMAGE_NAME=kserve/modelmesh-controller-develop
 DEV_DEPS="$0 Dockerfile.develop go.mod go.sum .pre-commit-config.yaml"
+ENGINE=$1
```
njhill, line 22: may as well keep default behaviour here too
```suggestion
ENGINE=${1:-docker}
```

njhill (commented):
Thanks @anishasthana, this looks good to me. I thought `docker` aliased to `podman` anyhow though?

Should we also add `--format=docker` to the `build` command in the podman case?

anishasthana:
It only happens if you've actually set up the alias ahead of time. I hadn't :-)

>Should we also add --format=docker to the build command in the podman case?

I actually haven't used the `--format=docker` flag before -- never found the need to. There doesn't seem to be a major difference in the image specs.

anishasthana:
I spoke to one of the podman maintainers. She said that the `--format=docker` is unnecessary unless there's something very specific to the docker format that our image requires. The OCI format is standard now and should be the default -- any runtimes that are OCI-compliant will work with OCI images.

njhill:
> It only happens if you've actually set up the alias ahead of time. 

Ah interesting, the RHEL install I use had it already so I assumed it was configured by default.

> I spoke to one of the podman maintainers. She said that the `--format=docker` is unnecessary unless there's something very specific to the docker format that our image requires. The OCI format is standard now and should be the default -- any runtimes that are OCI-compliant will work with OCI images.

OK sure, I just recall that I'd had to add it at some point in the past, maybe was for something obscure.

njhill (approved):

PR #267, opened 2022-10-24
Feat: Add PVC storage support

### LINKED ISSUE #230: PVC storage support
Support for serving models from PVCs is an important capability, we need to have a robust approach. It differs from our other cloud storage abstractions since it must be configured at the pod level, which means PVCs can't be configured dynamically in the multi-model case without restarting pods.

Thoughts:
1. The controller will need to be aware of Predictors tied to pvc storage. We need to think about whether we update the multi-model serving runtime Deployments dynamically if there's a new Predictor targeted for that deployment which uses a PVC not currently mounted there, and whether we refresh the Deployment again if/when a particular PVC is no longer needed.
    - We could make this configurable with reasonable defaults, for example reasonable default behaviour would probably be to avoid restarting a Deployment to just to remove pvcs which are no longer needed.
    - We should add a ServingRuntime field allowing to specify pvcs that should _always_ be mounted to the given runtime, and possibly a list in the global config of PVCs which should be mounted to all (multi-model) runtimes. This will help with common use-cases where there's only one (or a small/fixed number) used to store all models.
    - This isn't needed for single-model runtimes for hopefully obvious reasons 
2. PVC can be one of our abstract storage types, the corresponding puller logic would just symlink the targeted file or dir from the mounted pvc volume into the model dir (which itself _may_ be subsequently symlinked by adapter logic). We can read-only mount all of the "current" pvcs into one area (e.g. `/pvc_mounts/<pvc-name>/`).
3. Generally it probably doesn't make sense to have a config entry in our storage secret for pvcs since they are already separately configured and all that should be needed is the `claimName`, i.e. Predictor would contain something like:

        storage:
          type: pvc
          parameters:
              claimName: myPvc
4. We could consider retaining specific fields in the CRD since it's a special case. This would also help with 1 and 3 above.

Related (but different): https://github.com/kserve/modelmesh-serving/issues/229

---

### Pull Requests:

- https://github.com/kserve/modelmesh-serving/pull/267
- https://github.com/kserve/modelmesh-runtime-adapter/pull/36

chinhuang007:
#### Motivation
Support PVC as a storage option, to address issue https://github.com/kserve/modelmesh-serving/issues/230

#### Modifications

- Add PVCs to deployment
- Add secret watch to ServingRuntime controller to update deployment
- Add logic to ServingRuntime controller to create PVCs for predictors if the global setting is true
- Create volumes and PVCs based on deployment for ServingRuntimes
- Update Predictor controller to validate for PVCs

#### Result

- PVCs are read from the storage secret and created for all ServingRuntimes
- PVCs are read from the predictor spec and created for all ServingRuntimes if the global setting is true

#### Related Issues

- https://github.com/kserve/modelmesh-serving/issues/230
- https://github.com/kserve/modelmesh-runtime-adapter/pull/36

---

Resolves #230

### DIFF
--- a/controllers/modelmesh/const.go
+++ b/controllers/modelmesh/const.go
@@ -29,6 +29,7 @@ const (
 	SocketVolume    = "domain-socket"
 
 	ConfigStorageMount = "storage-config"
+	PVCMount           = "pvc-mount"
 
 	//The name of the puller container
 	PullerContainerName = "puller"
@@ -45,11 +46,17 @@ const (
 	//The env variable puller uses to configure the config dir (secrets)
 	PullerEnvStorageConfigDir = "STORAGE_CONFIG_DIR"
 
+	//The env variable puller uses to configure the pvc dir (secrets)
+	PullerEnvPVCDir = "PVC_DIR"
+
 	//The puller default port number
 	PullerPortNumber = 8086
 
 	//The puller model mount path
 	PullerModelPath = "/models"
+
+	//The puller model PVC path
+	PullerPVCPath = "/pvc_mounts"
 
 	//The puller model config path
 	PullerConfigPath = "/storage-config"
--- a/controllers/modelmesh/const.go
+++ b/controllers/modelmesh/const.go
@@ -29,7 +29,6 @@ const (
 	SocketVolume    = "domain-socket"
 
 	ConfigStorageMount = "storage-config"
-	PVCMount           = "pvc-mount"
 
 	//The name of the puller container
 	PullerContainerName = "puller"
@@ -47,7 +46,7 @@ const (
 	PullerEnvStorageConfigDir = "STORAGE_CONFIG_DIR"
 
 	//The env variable puller uses to configure the pvc dir (secrets)
-	PullerEnvPVCDir = "PVC_DIR"
+	PullerEnvPVCDir = "PVC_MOUNTS_DIR"
 
 	//The puller default port number
 	PullerPortNumber = 8086
@@ -56,7 +55,7 @@ const (
 	PullerModelPath = "/models"
 
 	//The puller model PVC path
-	PullerPVCPath = "/pvc_mounts"
+	DefaultPVCMountsDir = "/pvc_mounts"
 
 	//The puller model config path
 	PullerConfigPath = "/storage-config"
--- a/controllers/modelmesh/puller.go
+++ b/controllers/modelmesh/puller.go
@@ -67,6 +67,9 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 			}, {
 				Name:  PullerEnvStorageConfigDir,
 				Value: PullerConfigPath,
+			}, {
+				Name:  PullerEnvPVCDir,
+				Value: PullerPVCPath,
 			},
 		},
 		Image:   pullerImage,
@@ -87,6 +90,11 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 			{
 				Name:      ConfigStorageMount,
 				MountPath: PullerConfigPath,
+				ReadOnly:  true,
+			},
+			{
+				Name:      PVCMount,
+				MountPath: PullerPVCPath,
 				ReadOnly:  true,
 			},
 		},
--- a/controllers/modelmesh/puller.go
+++ b/controllers/modelmesh/puller.go
@@ -14,6 +14,7 @@
 package modelmesh
 
 import (
+	"path/filepath"
 	"strconv"
 
 	kserveapi "github.com/kserve/kserve/pkg/apis/serving/v1alpha1"
@@ -25,14 +26,14 @@ import (
 
 var StorageSecretName string
 
-func addPullerTransform(rts *kserveapi.ServingRuntimeSpec, pullerImage string, pullerImageCommand []string, pullerResources *corev1.ResourceRequirements) func(*unstructured.Unstructured) error {
+func addPullerTransform(rts *kserveapi.ServingRuntimeSpec, pullerImage string, pullerImageCommand []string, pullerResources *corev1.ResourceRequirements, pvcs []string) func(*unstructured.Unstructured) error {
 	return func(resource *unstructured.Unstructured) error {
 		var deployment = &appsv1.Deployment{}
 		if err := scheme.Scheme.Convert(resource, deployment, nil); err != nil {
 			return err
 		}
 
-		err := addPullerSidecar(rts, deployment, pullerImage, pullerImageCommand, pullerResources)
+		err := addPullerSidecar(rts, deployment, pullerImage, pullerImageCommand, pullerResources, pvcs)
 		if err != nil {
 			return err
 		}
@@ -41,7 +42,7 @@ func addPullerTransform(rts *kserveapi.ServingRuntimeSpec, pullerImage string, p
 	}
 }
 
-func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Deployment, pullerImage string, pullerImageCommand []string, pullerResources *corev1.ResourceRequirements) error {
+func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Deployment, pullerImage string, pullerImageCommand []string, pullerResources *corev1.ResourceRequirements, pvcs []string) error {
 	endpoint, err := ValidateEndpoint(*rts.GrpcMultiModelManagementEndpoint)
 	if err != nil {
 		return err
@@ -69,7 +70,7 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 				Value: PullerConfigPath,
 			}, {
 				Name:  PullerEnvPVCDir,
-				Value: PullerPVCPath,
+				Value: DefaultPVCMountsDir,
 			},
 		},
 		Image:   pullerImage,
@@ -92,11 +93,6 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 				MountPath: PullerConfigPath,
 				ReadOnly:  true,
 			},
-			{
-				Name:      PVCMount,
-				MountPath: PullerPVCPath,
-				ReadOnly:  true,
-			},
 		},
 	}
 
@@ -104,6 +100,13 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 		cspec.VolumeMounts = append(cspec.VolumeMounts, corev1.VolumeMount{
 			Name:      udsVolMountName,
 			MountPath: udsParentPath,
+		})
+	}
+	for _, pvcName := range pvcs {
+		cspec.VolumeMounts = append(cspec.VolumeMounts, corev1.VolumeMount{
+			Name:      pvcName,
+			MountPath: DefaultPVCMountsDir + string(filepath.Separator) + pvcName,
+			ReadOnly:  true,
 		})
 	}
 
--- a/controllers/predictor_controller.go
+++ b/controllers/predictor_controller.go
@@ -240,26 +240,36 @@ func (pr *PredictorReconciler) validatePredictor(predictor *api.Predictor) strin
 		return "Only one of spec.schemaPath and spec.storage.schemaPath can be specified"
 	}
 
-	// PersistentVolumeClaim is checked against the storage secret
-	if storage.PersistentVolumeClaim != nil {
-		s := &corev1.Secret{}
-		if err := pr.Client.Get(context.TODO(), types.NamespacedName{
-			Name:      modelmesh.StorageSecretName,
-			Namespace: predictor.Namespace,
-		}, s); err != nil {
-			return "Could not get the storage secret"
-		}
-		var storageConfig map[string]string
-		for _, storageData := range s.Data {
-			if err := json.Unmarshal(storageData, &storageConfig); err != nil {
-				return "Could not parse storage configuration json"
+	if storage.StorageSpec.Parameters != nil {
+		parameters := *storage.StorageSpec.Parameters
+		if storageType, ok := parameters["type"]; ok && storageType == StoragePVCType {
+			s := &corev1.Secret{}
+			if err := pr.Client.Get(context.TODO(), types.NamespacedName{
+				Name:      modelmesh.StorageSecretName,
+				Namespace: predictor.Namespace,
+			}, s); err != nil {
+				return "Could not get the storage secret"
 			}
-			if storageConfig["type"] == StoragePVCType && storageConfig["name"] == storage.PersistentVolumeClaim.ClaimName {
-				return ""
+			var storageConfig map[string]string
+			found := false
+			for _, storageData := range s.Data {
+				if err := json.Unmarshal(storageData, &storageConfig); err != nil {
+					return "Could not parse storage configuration json"
+				}
+				if storageConfig["type"] == StoragePVCType && storageConfig["name"] == parameters["name"] {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return "PersistentVolumeClaim name is not in the storage config secret"
 			}
 		}
+	}
 
-		return "PersistentVolumeClaim name is not in the storage config secret"
+	// PersistentVolumeClaim is deprecated and was never supported
+	if storage.PersistentVolumeClaim != nil {
+		return "spec.storage.PersistentVolumeClaim is not supported"
 	}
 
 	// S3 is deprecated and can not be specified alongside the new storage fields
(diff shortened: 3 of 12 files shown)


controllers/modelmesh/const.go
```diff
@@ -29,6 +29,7 @@ const (
 	SocketVolume    = "domain-socket"
 
 	ConfigStorageMount = "storage-config"
+	PVCMount           = "pvc-mount"
```
njhill, line 32: I don't think this is needed.

controllers/modelmesh/const.go
```diff
@@ -45,12 +46,18 @@ const (
 	//The env variable puller uses to configure the config dir (secrets)
 	PullerEnvStorageConfigDir = "STORAGE_CONFIG_DIR"
 
+	//The env variable puller uses to configure the pvc dir (secrets)
+	PullerEnvPVCDir = "PVC_DIR"
```
njhill, line 50: Maybe rename to `PVC_MOUNTS_DIR`?

controllers/modelmesh/const.go
```diff
@@ -45,12 +46,18 @@ const (
...
+
 	//The puller default port number
 	PullerPortNumber = 8086
 
 	//The puller model mount path
 	PullerModelPath = "/models"
 
+	//The puller model PVC path
+	PullerPVCPath = "/pvc_mounts"
```
njhill, line 59: Could change the name of this field to something like `DefaultPVCMountsDir`? .. i.e. default for the value above.

Just having it be a constant and no env var is fine too, I don't think we'll need to change it.

controllers/modelmesh/puller.go
```diff
@@ -67,6 +67,9 @@ func addPullerSidecar(rts *kserveapi.ServingRuntimeSpec, deployment *appsv1.Depl
 			}, {
 				Name:  PullerEnvStorageConfigDir,
 				Value: PullerConfigPath,
+			}, {
+				Name:  PullerEnvPVCDir,
+				Value: PullerPVCPath,
```
njhill, line 72: I don't think this is needed

controllers/servingruntime_controller.go
```diff
@@ -75,6 +80,10 @@ type ServingRuntimeReconciler struct {
 	ClusterScope bool
 	// whether the controller is enabled to read and watch ClusterServingRuntimes
 	EnableCSRWatch bool
+	// whether the controller is enabled to read and watch secrets
+	EnableSecretWatch bool
+	// whether the controller is enabled to deploy pvc mounts based on predictors
+	DeployPVCForPredictor bool
```
njhill, line 86: I was thinking that this would be a dynamic config parameter, added to the global configmap.

controllers/servingruntime_controller.go
```diff
@@ -224,6 +233,42 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			isvc := &list.Items[i]
+			if isvc.Spec.Predictor.Model != nil {
+				storageUri := isvc.Spec.Predictor.Model.PredictorExtensionSpec.StorageURI
+				if u, urlErr := url.Parse(*storageUri); urlErr == nil {
+					pvcs[u.Host] = struct{}{}
+				}
+			}
+		}
+	}
+
```
njhill, line 271: Would be good to move all the above into a separate func which returns the list of PVCs.

controllers/servingruntime_controller.go
```diff
@@ -224,6 +233,42 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		for i := range list.Items {
+			isvc := &list.Items[i]
+			if isvc.Spec.Predictor.Model != nil {
+				storageUri := isvc.Spec.Predictor.Model.PredictorExtensionSpec.StorageURI
+				if u, urlErr := url.Parse(*storageUri); urlErr == nil {
+					pvcs[u.Host] = struct{}{}
+				}
+			}
+		}
```
njhill, line 269: As discussed, change this to be a variation of [this](https://github.com/kserve/modelmesh-serving/blob/main/controllers/servingruntime_controller.go#L383) logic ... where the passed-in func is used to collect the list of PVCs. We can actually also keep the `runtimeSupportsPredictor` condition in there ... so that we only consider predictors that can actually land on this runtime.

main.go
```diff
@@ -357,6 +360,33 @@ func main() {
 	}
 	enableCSRWatch := checkCSRVar(EnableClusterServingRuntimeEnvVar, "ClusterServingRuntime", &v1alpha1.ClusterServingRuntime{})
 
+	checkSecretVar := func(envVar string, resourceName string, resourceObject client.Object) bool {
+		// default is true
+		envVarVal, _ := os.LookupEnv(envVar)
+		if envVarVal != FalseString {
+			err = cl.Get(context.Background(), client.ObjectKey{Name: "foo", Namespace: ControllerNamespace}, resourceObject)
```
njhill, line 367: Might be best to just check doing a `Get` on a secret called `storage-config`

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -318,7 +324,11 @@ func (isvcr InferenceServiceRegistry) Get(ctx context.Context, nname types.Names
 	p.Spec.Storage.SchemaPath = schemaPath
 	p.Spec.Storage.Parameters = &parameters
 	p.Spec.Storage.StorageKey = secretKey
-
+	if parameters["type"] == pvcStorageType {
+		p.Spec.Storage.PersistentVolumeClaim = &corev1.PersistentVolumeClaimVolumeSource{}
+		p.Spec.Storage.PersistentVolumeClaim.ClaimName = parameters["name"]
+		p.Spec.Storage.PersistentVolumeClaim.ReadOnly = true
+	}
```
njhill, line 331: Remove this per previous comment

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -162,6 +164,10 @@ func processInferenceServiceStorage(inferenceService *v1beta1.InferenceService,
 		}
 
 		switch u.Scheme {
+		case pvcStorageType:
+			modelPath = strings.TrimPrefix(u.Path, "/")
+			uriParameters["type"] = pvcStorageType
+			uriParameters["name"] = u.Host
```
njhill, line 170: :+1:

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -39,6 +40,7 @@ const (
 	runtimeAnnotation    = "serving.kserve.io/servingRuntime"
 
 	azureBlobHostSuffix = "blob.core.windows.net"
+	pvcStorageType      = "pvc"
```
njhill, line 43: I don't think a constant is needed here, best to keep it consistent with the others.

controllers/servingruntime_controller.go
```diff
@@ -623,3 +676,16 @@ func (r *ServingRuntimeReconciler) clusterServingRuntimeRequests(csr *kserveapi.
 
 	return requests
 }
+
+func (r *ServingRuntimeReconciler) secretRequests(secret *corev1.Secret, namespace string) []reconcile.Request {
```
njhill, line 680: Rename to `storageSecretRequests`? (otherwise a bit generic)

Also don't need a separate `namespace` arg, can just get it from the secret inside this func.

njhill (commented):
Thanks @chinhuang007, please see inline comments.

chinhuang007:
@njhill thanks for the detailed review as always! Just made a few changes so please take a look again.

controllers/servingruntime_controller.go
```diff
@@ -286,6 +301,50 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if err := json.Unmarshal(storageData, &storageConfig); err != nil {
+			return nil, fmt.Errorf("Could not parse storage configuration json: %w", err)
+		}
+		if storageConfig["type"] == StoragePVCType {
+			pvcsMap[storageConfig["name"]] = struct{}{}
+		}
+	}
+	// add pvcs for predictors when the global config is enabled
+	if r.DeployPVCForPredictor {
+		list := &v1beta1.InferenceServiceList{}
```
njhill, line 325: I think we still need to change this per [prior comment](https://github.com/kserve/modelmesh-serving/pull/267#discussion_r1020613091), to support different predictor "sources" (could be predictor or inferencservice). I will aim to write an example of what I mean.

controllers/predictor_controller.go
```diff
@@ -239,6 +240,33 @@ func validatePredictor(predictor *api.Predictor) string {
...
+			}, s); err != nil {
+				return "Could not get the storage secret"
+			}
+			var storageConfig map[string]string
+			found := false
+			for _, storageData := range s.Data {
+				if err := json.Unmarshal(storageData, &storageConfig); err != nil {
+					return "Could not parse storage configuration json"
+				}
+				if storageConfig["type"] == StoragePVCType && storageConfig["name"] == parameters["name"] {
```
njhill, line 259: We should check for the existence of the "name" key in both cases... with a wrong configuration this could panic.

controllers/predictor_controller.go
```diff
@@ -239,6 +240,33 @@ func validatePredictor(predictor *api.Predictor) string {
 		return "Only one of spec.schemaPath and spec.storage.schemaPath can be specified"
 	}
 
+	if storage.StorageSpec.Parameters != nil {
```
njhill, line 243: I actually think we can probably omit this whole section. If the PVC isn't present it will cause the predictor load to fail (with an informative message from the puller).

main.go
```diff
@@ -411,16 +433,18 @@ func main() {
...
+		Log:                   ctrl.Log.WithName("controllers").WithName("ServingRuntime"),
+		Scheme:                mgr.GetScheme(),
+		ConfigProvider:        cp,
+		ConfigMapName:         types.NamespacedName{Namespace: ControllerNamespace, Name: UserConfigMapName},
+		ControllerNamespace:   ControllerNamespace,
+		ControllerName:        controllerDeploymentName,
+		ClusterScope:          clusterScopeMode,
+		EnableCSRWatch:        enableCSRWatch,
+		EnableSecretWatch:     enableSecretWatch,
+		DeployPVCForPredictor: conf.EnableDeployPVCForPredictor,
```
njhill, line 446: Getting the value here means it won't be dynamic. We can just grab it from the config where we use it.

controllers/servingruntime_controller.go
```diff
@@ -286,6 +301,50 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	for _, storageData := range s.Data {
+		if err := json.Unmarshal(storageData, &storageConfig); err != nil {
+			return nil, fmt.Errorf("Could not parse storage configuration json: %w", err)
+		}
+		if storageConfig["type"] == StoragePVCType {
+			pvcsMap[storageConfig["name"]] = struct{}{}
+		}
+	}
+	// add pvcs for predictors when the global config is enabled
+	if r.DeployPVCForPredictor {
```
njhill, line 324: Per other comment this doesn't need to be a reconciler field, just pass `cfg` into this function and get the value from there.

pkg/config/config.go
```diff
@@ -55,8 +55,9 @@ var (
 // Config holds process global configuration information
 type Config struct {
 	// System config
-	EtcdSecretName    string // DEPRECATED - should be removed in the future
-	ModelMeshEndpoint string // For dev use only
+	EtcdSecretName              string // DEPRECATED - should be removed in the future
+	ModelMeshEndpoint           string // For dev use only
+	EnableDeployPVCForPredictor bool
```
njhill, line 60: I think we should name this differently, how about `AllowAnyPVC`?

We should also add to the docs ... the config params table and also in the storage part of the docs explain how this parameter works.

njhill (commented):
Thanks @chinhuang007 it's getting closer :)

One of the things is nontrivial and I'll aim to add some more detail for that.

controllers/servingruntime_controller.go
njhill: @chinhuang007 here is what I was thinking of, sorry again for the delay:
```go
	if cfg.AllowAnyPVC {
		restProxyEnabled := cfg.RESTProxy.Enabled
		f := func(p *api.Predictor) bool {
			if runtimeSupportsPredictor(rt, p, restProxyEnabled, req.Name) &&
				p.Spec.Storage != nil && p.Spec.Storage.Parameters != nil {
				
				params := *p.Spec.Storage.Parameters
				if stype := params["type"]; stype == "pvc" {
					if name := params["name"]; name != "" {
						pvcsMap[name] = struct{}{}
					}
				}
			}
			return false
		}

		for _, pr := range r.RegistryMap {
			if found, err := pr.Find(ctx, req.Namespace, f); found || err != nil {
				return nil, err
			}
		}
	}
```

controllers/servingruntime_controller.go
```diff
@@ -286,6 +299,57 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	var storageConfig map[string]string
+	for _, storageData := range s.Data {
+		if err := json.Unmarshal(storageData, &storageConfig); err != nil {
+			return nil, fmt.Errorf("Could not parse storage configuration json: %w", err)
+		}
+		if storageConfig["type"] == StoragePVCType {
+			if _, exists := storageConfig["name"]; !exists {
+				return nil, fmt.Errorf("Missing PVC name in storage configuration")
+			}
+			pvcsMap[storageConfig["name"]] = struct{}{}
```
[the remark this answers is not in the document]
njhill: Actually on second thoughts, simplest would just be this. I also think we shouldn't return a failure for this, best to just log a warning.
```suggestion
			if name := storageConfig["name"]; name != "" {
				pvcsMap[name] = struct{}{}
			} else {
				// Log warning here
			}
```

pkg/predictor_source/inferenceservice_registry.go
```diff
@@ -335,8 +335,23 @@ func (isvcr InferenceServiceRegistry) Find(ctx context.Context, namespace string
 	}
 
 	for i := range list.Items {
-		p, _ := BuildBasePredictorFromInferenceService(&list.Items[i])
-		if p != nil && predicate(p) {
+		inferenceService := &list.Items[i]
+		nname := types.NamespacedName{Name: inferenceService.Name, Namespace: inferenceService.Namespace}
+		p, err := BuildBasePredictorFromInferenceService(inferenceService)
```
njhill, line 340: @chinhuang007 we should probably check for `p == nil` here like before.

fOO223Fr:
@pvaneck @ckadner is there any update on this Feature? this would be a great addition for our use case. Thank you

njhill:
@fOO223Fr yes this is the next thing we are focused on completing. It didn't make the 0.10 release but is mostly complete. The main thing remaining to do is some tests, which @ckadner is working on.

controllers/servingruntime_controller.go
```diff
@@ -286,6 +298,60 @@ func (r *ServingRuntimeReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 }
 
+func (r *ServingRuntimeReconciler) getPVCs(ctx context.Context, req ctrl.Request, rt *kserveapi.ServingRuntimeSpec, cfg *config.Config) ([]string, error) {
+	s := &corev1.Secret{}
+	if err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      modelmesh.StorageSecretName,
+		Namespace: req.Namespace,
+	}, s); err != nil {
+		return nil, fmt.Errorf("Could not get the storage secret: %w", err)
+	}
```
Jooho, line 308: To use PVC, why do we need to check StorageSecretName The default "storage-config" secret usually has credentials for accessing cloud storage such as S3. However, PVC does not need credentials to attach to the pod.
With this code block, the controller keeps checking if the secret exists in a namespace. So although it is not needed, it must exist in the namespace to deploy an inference pod.

ckadner: This goes back to some of the design decisions for the implementation, i.e. ...

https://github.com/kserve/modelmesh-serving/issues/230#issuecomment-1260136468

> The controller will watch the storage config secret in each namespace and keep the pvcs mounted to runtime deployments based on any pvc type entries
Predictors can always reference one of these PVCs for storage

https://github.com/kserve/modelmesh-serving/issues/230#issuecomment-1277980357

> The entry type will be checked and the name used to create volumes and volume mounts for runtime deployments, while the keys are irrelevant.

Jooho: so even though users want to use PVC only, creating `storage-config` Secret with dummy data is required, am I right?

ckadner: If you use the `allowAnyPVC` config parameter, and don't want to deploy the minio-examples image, I suppose you would not need it 
Certainly it should not be required to create it with dummy data, so I think we should tolerate it not being there. @njhill

njhill: @ckadner yes as discussed offline we should change it so that in that mode the storage-config secret isn't required.

njhill (approved):
Thanks @chinhuang007 @ckadner!

PR #272, opened 2022-10-28
feat: Enable HA by default for Modelmesh Controller

anishasthana:
The pod anti-affinity will tell the Kubernetes schedule to schedule pods on different nodes if possible. This will allow users to enable HA by setting the number of replicas to a number greater than 1.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -21,7 +21,7 @@ spec:
   selector:
     matchLabels:
       control-plane: modelmesh-controller
-  replicas: 1
+  replicas: 1 # This can be increased safely to enable HA. A good value to set is 3.
   template:
     metadata:
       labels:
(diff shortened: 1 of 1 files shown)


anishasthana:
cc @njhill

njhill:
@anishasthana sorry for the delay. I'm unsure of whether we want to do this by default since the controller doesn't really require HA. If it is restarted / temporarily down there should not be much noticeable impact to the running services since its only part of the control plane and the api server will still accept requests.

Do you know whether it's typical practice for kubernetes controllers to have multiple replicas, e.g. in OpenShift?

I'm fine with adding the anti-affinity rule, and users are free to increase the number of replicas, just not sure of setting the default to > 1 (and thereby increasing the baseline footprint).

anishasthana:
@njhill makes sense! I've added the pod anti-affinity as well as a comment letting users know it's safe to increase the replicas.

I don't know if it's "typical" practice, but I know a lot of SRE teams and the like greatly prefer having multiple pods where possible to minimize impact, even if the impact itself is barely noticeable.

njhill (approved):
Thanks @anishasthana

PR #289, opened 2022-12-10
fix: config-defaults mount path

### LINKED ISSUE #205: Manager config mounted wrongly
**To Reproduce**

1. install modelmesh in microk8s from scratch without --quickstart
2. use `kubectl exec` to get in the manager pod
3. cd /etc/model-serving
4. the config-defaults.yaml here is a directory, which contains a real config-defaults.yaml
5. the mount setting can be found in https://github.com/kserve/modelmesh-serving/blob/main/config/manager/manager.yaml

**Expected behavior**

the config-defaults.yaml should not be a directory itself

ckadner:
#### Motivation

The file containing the default configuration for the ModelMesh Controller is called `config-defaults.yaml`, which resides in a directory also called `config-defaults.yaml` as pointed out by @adrwong-lora in issue #205
```
/etc/model-serving/config-defaults.yaml/config-defaults.yaml
```

#### Modifications

Change `mountPath` from `/etc/model-serving/config-defaults.yaml` to `/etc/model-serving/config/default`

#### Result

The file containing the default configuration file after this fix:

```
/etc/model-serving/config/default/config-defaults.yaml
```

Which more closely reflects the path of the [`config-defaults.yaml`](https://github.com/kserve/modelmesh-serving/blob/main/config/default/config-defaults.yaml) file in the `modelmesh-serving` repository:
```
config/default/config-defaults.yaml
```

Resolves #205 

/cc @njhill 
/fyi @adrwong-lora

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -81,7 +81,7 @@ spec:
               cpu: "1"
               memory: 512Mi
           volumeMounts:
-            - mountPath: /etc/model-serving/config-defaults.yaml
+            - mountPath: /etc/model-serving/config/default
               name: config-defaults
               readOnly: true
           securityContext:
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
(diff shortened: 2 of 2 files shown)


njhill (approved):
Thanks @ckadner

PR #304, opened 2023-01-10
Use `--namespace-scope-mode` for Quickstart

ckadner:
#### Motivation

Quickstart should use namespace scope mode for simplicity. In current version there is an inconsistency wrt `ServingRuntimes` vs `ClusterServingRuntimes`.

#### Modifications

- `docs/install/install-script.md` (correct a typo)
- `docs/quickstart.md` (change install command for ns scope mode and add missing `torchserv` serving runtime)
- `scripts/install.sh` (fix `sed -i` commands)

#### Result

Quickstart install steps work like a charm again :-)

#### TODOs

Once merged into `main`, we will need to update the `release-0.10` branch as well

---

/cc @njhill 
/assign @njhill


njhill (approved):
Thanks @ckadner !

PR #306, opened 2023-01-10
doc: include namespace-scope info

rafvasq:
#### Motivation
There doesn't seem to be much information about what the `--namespace-scope-mode` flag does, or the difference between cluster-scoped and namespace-scoped serving runtimes.

#### Modifications
- As a start, a blurb is included for the `--namespace-scope-mode` flag in the `install-script` doc.
- Corrects the `--help` output as well.

#### Result
Helpful information about the `--namespace-scope-mode` which is included in the recommended way to use the quick-start install of ModelMesh.


njhill (approved):
Thanks @rafvasq!

PR #326, opened 2023-02-10
fix: update links to env var lines

rafvasq:
#### Motivation
Links pointing to ModelMesh environment variables are no longer accurate.

#### Modifications
Update the line numbers to point to the correct places in [ModelMeshEnvVars.java](https://github.com/kserve/modelmesh/blob/main/src/main/java/com/ibm/watson/modelmesh/ModelMeshEnvVars.java)

### DIFF
--- a/controllers/modelmesh/modelmesh.go
+++ b/controllers/modelmesh/modelmesh.go
@@ -258,24 +258,24 @@ func (m *Deployment) addMMEnvVars(deployment *appsv1.Deployment) error {
 	}
 
 	if m.EnableAccessLogging {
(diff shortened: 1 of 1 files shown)


ckadner:
Maybe we should link to a release tag so the links don't go wonder of after changes to `main`. They would still be at risk of becoming obsolete if future versions deprecate/remove any of them but that would cause compile warning/errors at which point we could update the links.

rafvasq:
Thanks @ckadner, I agree. I updated the links point to v0.10.0.

ckadner (approved):
/lgtm

njhill (approved):

PR #339, opened 2023-03-06
Add support for configurable MM_Payload_Processor environment variable

RobGeada:
#### Motivation
In order to integrate with the upcoming [ModelMesh payload processing feature](https://github.com/kserve/modelmesh/pull/84#discussion_r1110428472), the MM_PAYLOAD_PROCESSOR environment variable within the ModelMesh image needs to point at the endpoint of the payload processing service within the cluster. By setting this as a variable set in the config.yaml, it allows for a variety of payload processing or logging services to be plugged in via the deployment manifest.

#### Modifications

- Added PayloadProcessor field to the `controllers/mmesh/modelmesh.go` Deployment struct
- Added PayloadProcessor field to the `pkg/config/config.go` Config struct
- Added default setting of this field to an empty string within `pkg/config/config.go` 
- Added functions in `controllers/suite_text.go` to load a config file that specifies the Payload processor
    - these tests ensure the default value of `MM_PAYLOAD_PROCESSOR` is empty, while it receives the correct value if such a field is present in the config yaml file 

#### Result
- Ability to set MM_PAYLOAD_PROCESSOR env variable via config.yaml
- If no such variable is set, the only effect is the ModelMesh image receives an empty env var


config/internal/base/deployment.yaml.tmpl
```diff
@@ -60,6 +60,9 @@ spec:
             - name: MM_SERVICE_NAME
               value: {{.ServiceName}}
               # External gRPC port of the service, should match ports.containerPort
+            - name: MM_PAYLOAD_PROCESSORS
+              value: {{.PayloadProcessors}}
+              # External endpoint for a payload processing service
```
njhill, line 65: This comment should be moved above the corresponding env var I think?

config/internal/base/deployment.yaml.tmpl
```diff
@@ -60,6 +60,9 @@ spec:
             - name: MM_SERVICE_NAME
               value: {{.ServiceName}}
               # External gRPC port of the service, should match ports.containerPort
+            - name: MM_PAYLOAD_PROCESSORS
+              value: {{.PayloadProcessors}}
```
njhill, line 64: Since this sometimes may be empty, might be best to surround the whole name/value with an `{{if ...` like what's done for `MM_METRICS` below.

Alternatively we could add it as a conditional step in the [addMMEnvVars](https://github.com/kserve/modelmesh-serving/blob/main/controllers/modelmesh/modelmesh.go#L215) function.

I'm not sure to be honest why some of the env vars get set via the template and others programmatically, I think it's just how the code evolved.

controllers/modelmesh/modelmesh.go
```diff
@@ -47,6 +47,7 @@ type Deployment struct {
 	Metrics            bool
 	PrometheusPort     uint16
 	PrometheusScheme   string
+	PayloadProcessors  string
```
njhill, line 50: I think it would be better for this to be a `string[]`, and then validate that none of the values contains a space.

njhill (commented):
Thanks @RobGeada, just a couple of comments

controllers/modelmesh/modelmesh.go
RobGeada: @njhill would the plan be then:

1) Make the modelmesh.PayloadProcessors a `string[]`
2) Have the `config.yaml `schema expect a list in the `PayloadProcessors` yaml field
3) When the `modelmesh.PayloadProcessors` struct is used to construct the `pkg/config` struct (i.e., [here](https://github.com/kserve/modelmesh-serving/pull/339/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R339), unpack the array, check if there's no spaces in any item, and pass it to the config struct as a space-separated string?

njhill: @RobGeada yes, exactly.

Kind of separate though but I do have a slight concern that the current URI spec for the processors might need more thought to ensure it's sufficiently generic, for example I imagine it may be common to want to enable for _all_ models not just specific ones, etc. But that doesn't really affect this PR.

RobGeada: Sorry, hang on, shouldn't it be the other way around?

The `pkg/config` struct creates the `modelmesh.Deployment` struct, which is then used to populate the deployment template. Surely we want the `pkg/config` to accept the string[] from the yaml, and then pass it to the space separated string in the `modelmesh.Deployment` struct.

So:

- [pkg/config](https://github.com/RobGeada/modelmesh-serving/blob/ab45d5cbca172e28fcba31e18507da80071543c3/pkg/config/config.go#L79).PayloadProcessors -> string[], from yaml
- [controllers/modelmesh.Deployment.PayloadProcessors](https://github.com/RobGeada/modelmesh-serving/blob/PayloadProcessorEndpointConfig/controllers/modelmesh/modelmesh.go) -> string
- In [servingruntime_controller](https://github.com/RobGeada/modelmesh-serving/blob/ab45d5cbca172e28fcba31e18507da80071543c3/controllers/servingruntime_controller.go#L251), parse `pkg/config` string array into a space separated list

RobGeada:
@njhill, comments addressed!

config/internal/base/deployment.yaml.tmpl
```diff
@@ -59,10 +59,15 @@ spec:
...
-              # External gRPC port of the service, should match ports.containerPort
+            # External endpoint for a payload processing service
+            {{if .PayloadProcessors}}
             - name: MM_PAYLOAD_PROCESSORS
               value: {{.PayloadProcessors}}
-              # External endpoint for a payload processing service
+            {{- else }}
+            - name: MM_PAYLOAD_PROCESSORS
+              value:
+            {{end}}
```
njhill, line 69: Actually the reason I was suggesting this change was to avoid this... so that the env var is not set at all when no payload processors are configured. Could we remove the else block?

njhill (commented):
Thanks @RobGeada, see inline comment, and also you need to add `Signed-off-by` to your latest commit so that the DCO check passes.

config/internal/base/deployment.yaml.tmpl
RobGeada: So I tried that, but removing the else block doing that breaks the entire config; the test suite complains that it cannot find any of the yaml fields after the {{end}}

RobGeada: Here are the results of `make run test`, keeping everything else identical, but just deleting the else block:

[error.log](https://github.com/kserve/modelmesh-serving/files/10981578/error.log)

njhill: @RobGeada might it be something to do with the fact that `.PayloadProcessors` isn't a boolean and so a slightly different `{{if ..` is needed? Or that the .golden file needs to be updated to remove the env var name in the absent cases?

Failing that, maybe would be better to just add it to `[addMMEnvVars](https://github.com/kserve/modelmesh-serving/blob/main/controllers/modelmesh/modelmesh.go#L215)` instead as suggested in prior comment?

RobGeada: Okay @njhill, done

njhill (approved):
Thanks @RobGeada, LGTM!

PR #363, opened 2023-05-01
fix: Generalizes sed and fixes dry-run cmd

rafvasq:
#### Motivation
I experienced a few errors when running the install script using the `--enable-self-signed-ca` flag from the recently merged #342. 

#### Modifications
- Use `sed -i.bak` to work around different versions of it, originally resulting in `sed: -i: No such file or directory` for me.
- Fix `--dry-run` error which requires a set value [(none, server, or client)](https://stackoverflow.com/questions/69531858/what-is-the-different-dry-run-opportunities)
- Fix typos

#### Result
Fixes errors associated with the `--enable-self-signed-ca` flag

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -325,16 +325,17 @@ if [[ $enable_self_signed_ca == "true" ]]; then
   # comment out vars
   configMapGeneratorStartLine=$(grep -n configMapGenerator ./default/kustomization.yaml |cut -d':' -f1)
   configMapGeneratorBeforeLine=$((configMapGeneratorStartLine-1))
-  sed "1,${configMapGeneratorBeforeLine}s/^/#/g" -i default/kustomization.yaml
-  
+  sed -i.bak "1,${configMapGeneratorBeforeLine}s/^/#/g" default/kustomization.yaml
+
   # remove webhookcainjection_patch.yaml
-  sed 's+- webhookcainjection_patch.yaml++g' -i default/kustomization.yaml
+  sed -i.bak  's+- webhookcainjection_patch.yaml++g' default/kustomization.yaml
 
   # create dummy secret 'modelmesh-webhook-server-cert'
   secretExist=$(kubectl get secret modelmesh-webhook-server-cert --ignore-not-found|wc -l)
-  if [[ $secretExist == 0 ]]; then
+  if [[ $secretExist -eq 0 ]]; then
     kubectl create secret generic modelmesh-webhook-server-cert 
   fi
+  rm default/kustomization.yaml.bak
 fi
 
 kustomize build default | kubectl apply -f -
@@ -365,10 +366,9 @@ if [[ $enable_self_signed_ca == "true" ]]; then
   info "Enabled Self Signed CA: Generate certificates and restart controller"
   
   # Delete dummy secret for webhook server
-  kubectl delete secret modelmesh-webhook-server-cert 
+  kubectl delete secret modelmesh-webhook-server-cert
 
   ../scripts/self-signed-ca.sh --namespace $namespace
-
 fi
 
 info "Waiting for ModelMesh Serving controller pod to be up..."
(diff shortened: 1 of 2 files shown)


ckadner (approved):
/lgtm

ckadner:
Thanks for this fix @rafvasq -- Looks good. Any reason this is marked as a draft?

rafvasq:
> Thanks for this fix @rafvasq -- Looks good. Any reason this is marked as a draft?

Thanks @ckadner. I had marked it as draft because there was another potential bug I wanted to work out and include in this PR, but we could just open up another PR when we have the fix for that. Marking this ready for now.

ckadner:
@Jooho -- the other error

```
webhook-server.default": failed to call webhook: Post "https://modelmesh-webhook-server-service.modelmesh-serving.svc:9443/validate-serving-modelmesh-io-v1alpha1-servingruntime?timeout=10s": EOF
```

scripts/install.sh
```diff
@@ -325,16 +325,17 @@ if [[ $enable_self_signed_ca == "true" ]]; then
   # comment out vars
   configMapGeneratorStartLine=$(grep -n configMapGenerator ./default/kustomization.yaml |cut -d':' -f1)
   configMapGeneratorBeforeLine=$((configMapGeneratorStartLine-1))
-  sed "1,${configMapGeneratorBeforeLine}s/^/#/g" -i default/kustomization.yaml
-  
+  sed -i.bak "1,${configMapGeneratorBeforeLine}s/^/#/g" default/kustomization.yaml
+
   # remove webhookcainjection_patch.yaml
-  sed 's+- webhookcainjection_patch.yaml++g' -i default/kustomization.yaml
+  sed -i.bak  's+- webhookcainjection_patch.yaml++g' default/kustomization.yaml
```
tjohnson31415, line 331: Another fix we might need here: It looks like this line is attempting to remove the line
```
  - webhookcainjection_patch.yaml
```
but in the current version of `default/kustomization.yaml` the line is now 
```
  - path: webhookcainjection_patch.yaml
```
[REF](https://github.com/kserve/modelmesh-serving/blob/dd0229fe52d6c90703eabeecfdee287dd3ac55fd/config/default/kustomization.yaml#L65)

🤔 
I think we can just change this to a sed invocation that deletes a matching line instead of a replacement:
```
sed -i.bak '/webhookcainjection_patch.yaml/d' default/kustomization.yaml
```

Jooho: Yes I agree with travis. We can use the sed command to remove the line.

Jooho: Just for reference, I commented that the -i.bak is for Mac os version option.

scripts/install.sh
```diff
@@ -325,16 +325,17 @@ if [[ $enable_self_signed_ca == "true" ]]; then
...
 
   # create dummy secret 'modelmesh-webhook-server-cert'
   secretExist=$(kubectl get secret modelmesh-webhook-server-cert --ignore-not-found|wc -l)
-  if [[ $secretExist == 0 ]]; then
+  if [[ $secretExist -eq 0 ]]; then
     kubectl create secret generic modelmesh-webhook-server-cert 
   fi
+  rm default/kustomization.yaml.bak
```
Jooho, line 338: why don't we add `*.bak` into .gitignore file?

ckadner: we could add `*.bak` to `.gitignore` (and `.dockerignore`) but we should not accumulate temp files in the project directory

rafvasq:
> @Jooho -- the other error
> 
> ```
> webhook-server.default": failed to call webhook: Post "https://modelmesh-webhook-server-service.modelmesh-serving.svc:9443/validate-serving-modelmesh-io-v1alpha1-servingruntime?timeout=10s": EOF
> ```

No longer experiencing this one! Controller is coming up with the webhook configured as expected :)

tjohnson31415 (approved):
LGTM

Issue #279, opened 2022-11-11
InferenceService status is not updated correctly


DaoDaoNoCode:
**Describe the bug**

If I have some InferenceService resource running on a ServingRuntime, and I delete the ServingRuntime, the status of the InferenceService is still `Ready`

Check the pod log:
```
{"level":"info","ts":1668185801.959513,"logger":"controllers.Predictor","msg":"InferenceService Status updated","namespacedName":"mesh-test/example-onnx-mnist","source":"InferenceService","newStatus":{"available":true,"transitionStatus":"UpToDate","activeModelState":"Loaded","targetModelState":"","lastFailureInfo":{"reason":"RuntimeUnhealthy","message":"Waiting for runtime Pod to become available","modelId":"example-onnx-mnist__isvc-82e2bf7ea4"},"httpEndpoint":"http://modelmesh-serving.mesh-test:8008/","grpcEndpoint":"grpc://modelmesh-serving.mesh-test:8033","totalCopies":1,"failedCopies":0}}
```

The `available` is still `true`

**To Reproduce**

1. Create a ServingRuntime
2. Create an InferenceService running on the server
3. Delete ServingRuntime
4. Check the condition of the InferenceService

**Expected behavior**
As soon as the ServingRuntime is deleted, the InferenceService should be unavailable.

njhill:
@DaoDaoNoCode this sounds like a plausible bug. Need to think some more about the best way to address it.

PR #394, opened 2023-06-09
Updated /controllers/predictor_controller.go

BasakUlker:
#### Motivation
[#279 ](https://github.com/kserve/modelmesh-serving/issues/279)
#### Modifications
Updated check order on "decodeModelState" function in /controllers/predictor_controller.go
#### Result
According to this [bug](https://github.com/kserve/modelmesh-serving/issues/279), I think the reason why InferenceService's state is not updated correctly is because the error is not defined correctly. If there is no ServingRuntime defined with the specified runtime name, reason should be: "RuntimeNotRecognize". The main cause of the error is "RuntimeNotRecognize", but it satisfies the conditions for the "RuntimeUnhealthy" error, and the error is defined as "RuntimeUnhealthy" because this check is done before the other error. The mentioned check is done in "decodeModelState" function in /controllers/predictor_controller.go. I made a small change in the function that I thought the error was detected, but I could not try the change. Because my computer has a shortage of ram and I'm having trouble with images. But if my idea is approved, I can solve my ram related issue and continue development.

### DIFF
--- a/controllers/predictor_controller.go
+++ b/controllers/predictor_controller.go
@@ -390,6 +390,7 @@ func concreteModelName(predictor *api.Predictor, sourceId string) string {
 // "There are no running instances that meet the label requirements of type _default: [_no_runtime]"
 const noHomeMessage string = "There are no running instances that meet the label requirements of type "
 
+//If this function setting the error, "RuntimeNotRecognize" should've to be controlled before "RuntimeUnhealthy". So, order changed.
 func decodeModelState(status *mmeshapi.ModelStatusInfo) (api.ModelState, api.FailureReason, string) {
 	reason := api.FailureReason("")
 	msg := ""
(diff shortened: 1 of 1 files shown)


rafvasq:
Hey @BasakUlker, thanks for trying to tackle this bug. I can spend more time investigating too but unfortunately your change doesn't solve it. The status is updated [here](https://github.com/kserve/modelmesh-serving/blob/main/controllers/predictor_controller.go#L165-L169).

rafvasq (commented):
Hey @BasakUlker, thanks for trying to tackle this bug. I can spend more time investigating too but unfortunately your change doesn't solve it. The status is updated [here](https://github.com/kserve/modelmesh-serving/blob/main/controllers/predictor_controller.go#L165-L169).

rafvasq:
Feel free to open another PR if you'd like to tackle the bug again!

Issue #316, opened 2023-01-18
Add some python client examples to the docs


njhill:
Some self-contained / end-to-end examples of:

- Using the python SDK to manipulate ServingRuntimes / InferenceServices
- Generating gRPC stubs for the KServe V2 API and using these to make inference requests, including some that demonstrate using string/bytes datatype
- Making inferencing requests using the KServe V2 REST API

rafvasq:
A couple of relevant resources re: Python SDK 
- https://kserve.github.io/website/0.10/sdk_docs/sdk_doc/
- https://github.com/kserve/modelmesh-serving/issues/293

lamoghaighoroje:
A good first issue for me to contribute in the community. So I will work on this now. Thanks

njhill:
Thanks @lamoghaighoroje!

PR #395, opened 2023-06-09
doc: Add Python Client Examples

lgdeloss:
#### Motivation

- There is lack of documentation for how to interface with modelmesh-serving using python sdk

#### Modifications

- Added documentation on how to interface with Python client sdk for modelmesh-serving
- Added  `docs/samples/modelmesh_sdk_sample.ipynb`

#### Result

- Works on https://github.com/kserve/modelmesh-serving/issues/316


rafvasq (changes requested):
Few more updates. I think there'd be some confusion on the headers and flow of the sample if you read it like a new user. You can probably be a bit more descriptive like "Check InferenceService status after deploying" or " etc.

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,597 @@
...
+       "  'namespace': 'modelmesh-serving',\n",
+       "  'resourceVersion': '3881193',\n",
+       "  'uid': '62b375b1-fd66-4783-b391-d4c3f8e8e6d4'},\n",
+       " 'spec': {'predictor': {'sklearn': {'name': '',\n",
+       "    'protocolVersion': 'v2',\n",
+       "    'storage': {'key': 'localMinIO', 'path': 'sklearn/mnist-svm.joblib'}},\n",
+       "   'tensorflow': {'name': '',\n",
+       "    'protocolVersion': 'v2',\n",
+       "    'storage': {'key': 'localMinIO',\n",
+       "     'path': 'gs://kfserving-examples/models/tensorflow/flowers-2'}}}},\n",
```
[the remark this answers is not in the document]
lgdeloss: @rafvasq I just re-ran and noticed its still showing this path. I deleted the isvc  and re-executed steps. Any idea on why thats not updating?

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,589 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "namespace = 'modelmesh-serving'\n",
+    "name='msnist-sample'\n",
```
rafvasq, line 50:
```suggestion
    "name='mnist-sample'\n",
```

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,589 @@
...
+       "    'manager': 'OpenAPI-Generator',\n",
+       "    'operation': 'Update',\n",
+       "    'time': '2023-06-09T21:41:54Z'}],\n",
+       "  'name': 'msnist-sample',\n",
+       "  'namespace': 'modelmesh-serving',\n",
+       "  'resourceVersion': '4014187',\n",
+       "  'uid': '4360f17b-3575-4b0d-a324-3ba536b55697'},\n",
+       " 'spec': {'predictor': {'sklearn': {'name': '',\n",
+       "    'protocolVersion': 'v2',\n",
+       "    'storage': {'key': 'localMinIO', 'path': 'sklearn/mnist-svm.joblib'}},\n",
```
rafvasq, line 410: I noticed that `sklearn/mnist-svm.joblib` still appears along with `tensorflow/mnist.savedmodel` below in the same spec. It seems like in this case you'd actually want to use the `replace` as explained [here](https://kserve.github.io/website/0.10/sdk_docs/docs/KServeClient/#replace), instead of `patch`.

rafvasq (changes requested):

rafvasq (dismissed):
Thanks @lgdeloss, I think this is a good starting point for samples showing how to manipulate InferenceServices using the Python SDK.

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Sample for ModelMesh Serving using the KServe Python SDK\n",
```
ckadner, line 8: I think this should be a more concise title, like ...
- "Python SDK for ModelMesh"
- "Python SDK Examples"
- "How to use the KServe Python SDK with ModelMesh"
-

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "\n",
+    "## This notebook shows how to use the KServe SDK to create, update, and delete an InferenceService\n",
```
ckadner, line 10: I don't think we need a subtitle here; normal text should be fine, if we have a descriptive title / page heading we don't need a subtitle

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "\n",
+    "- This sample assumes ModelMesh Serving was deployed using the [quickstart guide](https://github.com/kserve/modelmesh-serving/blob/main/docs/quickstart.md)."
```
ckadner, line 12: no need for a single bullet point. we are not listing something here

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "from kserve import KServeClient"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### The variables being set are `namespace`, `name`, `protocol_version`\n",
+    "- The `namespace` definition is where the InferenceService will be deployed to\n",
```
ckadner, line 38: InferenceService should also be in-line code like `InferenceService` otherwise spell it in two words "Inference Service" -- @rafvasq what do you think?

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "- `name` will be the name of the InferenceService\n",
+    "- For ModelMesh, the `v2` protocol must be used since it doesn't support the default `v1` protocol"
```
ckadner, line 40: we should probably link to the v1 and v2 protocol pages here

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "name='mnist-sample'\n",
+    "protocol_version='v2'"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Define InferenceService specifying ModelMesh as the deploymentMode"
```
ckadner, line 59: maybe we can add a line here to describe why that is important?

> While both the KServe controller and ModelMesh controller will reconcile InferenceService resources, …
the ModelMesh controller will only handle those InferenceServices with the [serving.kserve.io/deploymentMode](http://serving.kserve.io/deploymentMode): ModelMesh annotation. Otherwise, the KServe controller will handle reconciliation.
the KServe controller will not reconcile an InferenceService with the [serving.kserve.io/deploymentMode](http://serving.kserve.io/deploymentMode): ModelMesh annotation, and will defer under the assumption that the ModelMesh controller will handle it.

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "kserve.get(name, namespace=namespace)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check InferenceService status once its ready\n",
+    "- State should now be `Loaded`"
```
ckadner, line 234: don't use bullets when not listing several items

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "kserve.wait_isvc_ready(name, namespace=namespace)\n",
+    "kserve.get(name, namespace=namespace)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Replace InferenceService and point it to a different model"
```
ckadner, line 320: should "Replace" be "Update"?

docs/samples/modelmesh_sdk_sample.ipynb
```diff
@@ -0,0 +1,562 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "{'apiVersion': 'serving.kserve.io/v1beta1',\n",
```
ckadner, line 108: I don't like these very long JSON outputs. Could we update the code blocks (here and below) to capture the output objects/maps/JSON data and use the Python map or Python JSON libraries to only print relevant text or parts of the output we care about, like the deployment status, the name/id of created objects, etc.?

ckadner (changes requested):
**Thank you Luis** (and Rafael). It's great to finally get a few Python SDK examples.

Should we rename the file to "python_sdk_examples.ipynb" since this entire repo is dedicated to "modelmesh".

I'm not sure if "samples" is the best name for the folder. Samples imply a user can take something and use it with/witout modifications, like a model, a YAML file, a script. This notebook is more like a "how to" guide or tutorial. If we plan more SDK examples in the future maybe the folder should be "sdk" or "python-sdk" or "notebooks" or "examples", if we think we will have more in the future.

I have a few more comments/suggestions.

docs/samples/modelmesh_sdk_sample.ipynb
lgdeloss: @ckadner I'm not sure how to do this that you're suggesting here to compress the json outputs. Can you share an example?

docs/samples/modelmesh_sdk_sample.ipynb
rafvasq: Yup I agree. I think it helps to make it more obvious that we're referring to the custom resource and not anything else. @lgdeloss, this would apply all around including headers like [this one](https://github.com/kserve/modelmesh-serving/pull/395/files#diff-9db730d6dd336df2fba42635af7136fa0b5190874583c43bae658aa07429332eR57)

docs/samples/modelmesh_sdk_sample.ipynb
rafvasq: @lgdeloss noticed you used a header style for the change here, I think you can just use regular text for the quickstart guide line (i.e. no `#`s)

docs/samples/modelmesh_sdk_sample.ipynb
ckadner: like what you do here:

```Python
print("status: ", updated["status"]["modelStatus"]["states"]["activeModelState"])
```

```
status:  Loaded
```

Or even just:

```Python
print(updated["status"]["modelStatus"]["states"]["activeModelState"])
```

```
Loaded
```

And only the parts that changed, are new ... no need to repeatedly print the same model ID / name if that already was outputted in a cell above

docs/examples/python_sdk_examples.ipynb
```diff
@@ -0,0 +1,363 @@
...
+    "print(pending_state[\"status\"][\"modelStatus\"][\"states\"][\"activeModelState\"])"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check `InferenceService` status once its ready\n",
+    "#### State should now be `Loaded`"
```
rafvasq, line 174: Normal text, not a header/subtitle
```suggestion
    "It should now be in a `Loaded` state."
```

rafvasq (changes requested):

rafvasq (changes requested):

rafvasq (approved):

docs/examples/python_sdk_examples.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "print(deployed_isvc[\"status\"][\"modelStatus\"][\"states\"][\"activeModelState\"])"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check `InferenceService` status once its ready\n",
+    "It should now be in a `Loaded` state."
```
ckadner, line 174: Even though we mentioned it above, I think we could explain when "now" is:

"Once the serving runtime pods are started, the `InferenceService` should be in `Loaded` state."

docs/examples/python_sdk_examples.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "                          spec=updated_spec)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "replace_isvc = kserve.replace(name, updated_isvc, namespace=namespace)"
```
ckadner, line 249: "replaced_isvc"

docs/examples/python_sdk_examples.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "print(\"path: \", replace_isvc[\"spec\"][\"predictor\"][\"tensorflow\"][\"storage\"][\"path\"])"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "kserve.wait_isvc_ready(name, namespace=namespace)\n",
```
ckadner, line 277: Maybe we could add another text cell above this one to explain that/why we need to wait for isvc to be ready. Even though the method name is indicating it.

ckadner (changes requested):
This is looking good! Just a few more nit-picks.

ckadner (approved):
Thanks @lgdeloss 

/lgtm

Issue #385, opened 2023-05-31
Memory leak in go routine unmarshalling secrets


ckadner:
**Describe the bug**

Memory leak in go routine that unmarshalls secrets. Repprted by @Jooho after a load test with 1000 simulated users.

https://github.com/kserve/modelmesh-serving/blob/main/pkg/mmesh/etcdrangewatcher.go#L97-L196

**Analysis**

From @tjohnson31415:

> The EtcdRangeWatcher does not have a way to exit the go-routine: there is no break out of the "main" refresh_loop.
The next question is if the number of watches should be increasing. I'm looking at how these are created now.
> The watchers are created in [refreshWatches](https://github.com/kserve/modelmesh-serving/blob/0bd750ed623378c1d6b00a8e68314f3ff1e83cc1/pkg/mmesh/modelmesh_events.go#L156)
> Which is called from [UpdateWatchedService](https://github.com/kserve/modelmesh-serving/blob/0bd750ed623378c1d6b00a8e68314f3ff1e83cc1/pkg/mmesh/modelmesh_events.go#LL98C34-L98C54)
which has if checks in both cases that it calls `refreshWatches`:
> 1. `if etcdSecretName != mes.secretName`
> 2. `else if serviceName != nw.watchedServiceName`
> The second check with `serviceName` might be the problem (the `etcd` secret shouldn't be changing much). It looks like the `nw` objects are created without a `watchedServiceName` being set:
> ```
> 		nw = &namespaceWatch{}
> ```
> [REF](https://github.com/kserve/modelmesh-serving/blob/0bd750ed623378c1d6b00a8e68314f3ff1e83cc1/pkg/mmesh/modelmesh_events.go#LL110C1-L110C25)
> So `nw.watchedServiceName` might always be an empty string 🤔 

**Screenshots**

![image](https://github.com/kserve/modelmesh-serving/assets/12246093/2c113508-61b9-4a33-9e04-a71e69c989e3)

**Environment (please complete the following information):**

**Additional context**

TODO

ckadner:
TODO: break out separate issue for Secret unmarshalling

PR #397, opened 2023-06-13
fix: Goroutine memory leak

Jooho:
#### Motivation
The goroutines watching ETCD data are never released from memory until the process is down because the logic to stop the goroutines does not exist. It only deletes [Service](https://github.com/kserve/modelmesh-serving/blob/main/controllers/service_controller.go#L117-L141)

 Moreover, it tries to get data for a namespace even if the namespace does not exist. It causes a lot of garbage logs like the following
~~~
{"level":"error","ts":"2023-06-13T19:37:13Z","msg":"Reconciler error","controller":"ServiceReconciler","controllerGroup":"","controllerKind":"Namespace","Namespace":{"name":"openshift-build-test-28111211-5dttn"},"namespace":"","name":"openshift-build-test-28111211-5dttn","reconcileID":"4daa4200-cc10-4365-81e6-54840dffa332","error":"Namespace \"openshift-build-test-28111211-5dttn\" not found","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235"}
~~~
ISSUE: https://github.com/kserve/modelmesh-serving/issues/385
#### Modifications
- Goruntines will be stopped when the label `modelmesh-enabled` is removed from a namespace or the namespace is deleted. 
- If the request namespace does not exist, it will not retry to get the information again.

#### Result
When the lable `modelmesh-enabled` is added to a namespace, 7 Goroutines are created but if the label is removed or the namespace is removed, 7 Goroutines will be stopped. The Goroutines count can be checked with pprof.
Unfortunately, Modelmesh does not have pprof pkg so it is not possible to check it directly. 

In order to test this, pprof pkg needs to be added to main. This PR has [the code example](https://github.com/opendatahub-io/modelmesh-serving/pull/110/files). If Modelmesh needs this part, I will contribute this with another PR.


controllers/util.go
```diff
@@ -24,6 +24,10 @@ import (
 
 func modelMeshEnabled(n *corev1.Namespace, controllerNamespace string) bool {
 	if v, ok := n.Labels["modelmesh-enabled"]; ok {
+		// Returns false if the namespace state is terminating even though the namespace have the 'modelmesh-enabled=true' label.
```
ckadner, line 27: This comment is helpful. Could you add comments to each of the changed code blocks above to describe what the code does and why (what problem it is solving).

ckadner (commented):
Hi @Jooho -- this is my third attempt to fully grasp the problem and your code changes. Without adding 1000 namespaces and a memory profiler, do you have a reasonably easy way to test the code before and after. Even a manual setup with 2 or 3 namespaces and a few `kubectl` commands to verify the state of the cluster/MM deployment resources and/or log messages.

controllers/service_controller.go
```diff
@@ -119,24 +119,34 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		namespace = req.Name
 		n := &corev1.Namespace{}
 		if err := r.Client.Get(ctx, req.NamespacedName, n); err != nil {
+			if k8serr.IsNotFound(err) {
```
Jooho, line 122: Previously, controller keep checking namespaces even though the namespaces do not exist anymore. 
As a result,  a lot of misleading error messages showed up in the log. (I don't find the snippet of the message now)

controllers/service_controller.go
```diff
@@ -119,24 +119,34 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 				}
 			}
+
 			if mms := r.MMServices.Get(namespace); mms != nil {
 				mms.Disconnect()
 				r.MMServices.Delete(namespace)
+				//requeue is never expected here
+				if _, err, _ := r.reconcileService(ctx, mms, namespace, owner); err != nil {
+					return ctrl.Result{}, err
+				}
```
Jooho, line 146: If the namespace is not for modelmesh anymore, it should trigger reconcileService for MMService list that manages the goroutines.

controllers/service_controller.go
```diff
@@ -226,6 +236,18 @@ func (r *ServiceReconciler) reconcileService(ctx context.Context, mms *mmesh.MMS
...
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: namespace}, n); err != nil {
+		return nil, err, false
+	}
+
+	if !modelMeshEnabled(n, r.ControllerDeployment.Namespace) {
+		r.ModelEventStream.RemoveWatchedService(serviceName, namespace)
+		r.Log.Info("Deleted Watched Service", "name", serviceName, "namespace", namespace)
+		return nil, nil, false
+	}
+
```
Jooho, line 250: This will remove the goroutine when modelmesh is not enabled for a namespace.
- when the namespace does not have the annotation `modelmesh-enabled`
- when the namespace is under a `Terminating` state.

Jooho:
@ckadner added comments to the source.

tjohnson31415 (commented):
I was able to verify that these changes prevent a growth in the number of go routines when creating and deleting namespaces. To see the count of go routines, I used the pprof patch linked above.

The recent change to add more comments has caused the build to fail because the `make fmt` check doesn't pass.

Also, I'm not sure that the reconciler will always see the namespace as terminating, and I wonder if the "namespace terminating" logic would be covered by implementing cleanup when the namespace has been deleted. This would simplify the changes too.

controllers/service_controller.go
Jooho: @tjohnson31415 Thanks for the review.  From my 2 cents, the operator keeps watching the namespace so I don't think there is a possibility of missing the terminated status of the Namespace object. When any namespace changes its status, this operator detects it. WDYT?

Regarding the `IsNotFound` part, do you want me to update this pr without it? and send another pr to update it when this original PR is merged?

Jooho: one more comment about the namespace object status. The main issue was the operator does not have logic when the namespace is under `Terminating` status([source](https://github.com/kserve/modelmesh-serving/blob/main/controllers/util.go#L26)).  From my understanding,  `Terminating` is the only problematic status in this procedure
- `Active` + ` modelmesh-enabled"` label ==> modelMeshEnabled return true
- `Terminating` + ` modelmesh-enabled"` label ==> modelMeshEnabled return true
- `Active` + no label ==>  modelMeshEnabled return false
The other status does not impact this logic.

The most important point here is, that when any namespace's status changes, this [reconclie](https://github.com/kserve/modelmesh-serving/blob/main/controllers/service_controller.go#L112) will be called and check the above logic. When any namespace gets `terminate` signal, the status will be changed to `Terminating`. 

please correct me if something is wrong here.

tjohnson31415: In the "happy path" case, I think that reconciliation should always be able to observe the terminating state. In error/edge cases (eg. many namespaces deleted in bulk, error encountered during reconciliation and a requeue happening), I'm not so sure since "Terminating" is a temporary status.

Before this PR, the reconciliation handled neither "Terminating" nor "Not Found". This PR implements clean-up if the namespace is seen as Terminating, but, If reconciliation happens for a namespace that no longer exists, should the code:
(a) Assume that in-memory resources for that namespace do not exists and just skip reconciliation (i.e. assume that if modelmesh was enabled for that namespace that it previously reconciled while it was Terminating)
(b) Check for in-memory resources that exist and clean them up if they do

With the "level-based" reconciliation that Operators are meant to implement, I think (b) makes more sense as it reconciles based on the current state, without assumption of past reconciliation.

ckadner (approved):
Thanks @Jooho 

/lgtm

PR #421, opened 2023-08-22
fix: Skip `ClusterServingRuntime` deletion if deployed using `--namespace-scope-mode`

### LINKED ISSUE #420: `delete.sh` script `resource mapping not found` error
After deploying with `--namespace-scope-mode`, the provided delete script fails to clean everything up when it runs into a `resource mapping not found` error. 

---

> One more related aspect of the delete script not working properly:
> 
> When MM was installed according to the Quickstart guide with the `-namespace-scope-mode` option:
> 
> ```Bash
> ./scripts/install.sh --namespace-scope-mode --namespace modelmesh-serving --quickstart
> ```
> 
> then the `delete.sh` script will produce this error:
> 
> ```
> ...
> resource mapping not found for name: "mlserver-0.x" namespace: "" from "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
> ensure CRDs are installed first
> resource mapping not found for name: "ovms-1.x" namespace: "" from "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
> ensure CRDs are installed first
> resource mapping not found for name: "torchserve-0.x" namespace: "" from "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
> ensure CRDs are installed first
> resource mapping not found for name: "triton-2.x" namespace: "" from "STDIN": no matches for kind "ClusterServingRuntime" in version "serving.kserve.io/v1alpha1"
> ensure CRDs are installed first
> ```
> 
> And leave behind many resources in the `modelmesh-serving` namespace
> 
> This might warrant a separate issue, though.
>
> _Originally posted by @ckadner in https://github.com/kserve/modelmesh-serving/issues/223#issuecomment-1384786636_

rafvasq:
#### Motivation
`delete.sh` script attempts to delete `ClusterServingRuntime` resources when they don't exist after deploying with `--namespace-scope-mode`, interrupting the script and leaving multiple resources behind.

#### Modifications
Set and use `NAMESPACE_SCOPE` env variable to check whether to skip `ClusterServingRuntime` deletion.

#### Result
Resolves #420


scripts/delete.sh
```diff
@@ -131,9 +131,15 @@ if [[ "$crb_ns" == "$namespace" ]]; then
   echo "deleting cluster scope RBAC"
   kustomize build rbac/cluster-scope | kubectl delete -f - --ignore-not-found=true
 fi
+
+# Determine whether deployment is namespace-scoped before deleting runtime resources
+is_namespace_scoped=$(kubectl exec deploy/modelmesh-controller -- printenv NAMESPACE_SCOPE)
 kustomize build default | kubectl delete -f - --ignore-not-found=true
 kustomize build rbac/namespace-scope | kubectl delete -f - --ignore-not-found=true
-kustomize build runtimes ${kustomize_load_restrictor_arg} | kubectl delete -f - --ignore-not-found=true
+if [[ ! $is_namespace_scoped ]]; then
```
ckadner, line 139: If MM was installed with your updated `install.sh` script:

```Bash
kubectl set env deploy/modelmesh-controller NAMESPACE_SCOPE=false
```

then I think this test will not work, since the variable `$is_namespace_scoped` does exist (is defined) but the string value is "false"

ckadner (changes requested):
Hi @rafvasq -- I think the way we define and test for the `is_namespace_scoped` variable needs to be hardened a bit

scripts/delete.sh
```diff
@@ -131,9 +131,15 @@ if [[ "$crb_ns" == "$namespace" ]]; then
   echo "deleting cluster scope RBAC"
   kustomize build rbac/cluster-scope | kubectl delete -f - --ignore-not-found=true
 fi
+
+# Determine whether deployment is namespace-scoped before deleting runtime resources
+is_namespace_scoped=$(kubectl exec deploy/modelmesh-controller -- printenv NAMESPACE_SCOPE)
```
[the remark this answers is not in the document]
rafvasq: Fixed, good reminder to think about already-existing deployments when I'm making changes

scripts/delete.sh
rafvasq: Fixed

scripts/delete.sh
```diff
@@ -131,9 +131,15 @@ if [[ "$crb_ns" == "$namespace" ]]; then
...
   kustomize build rbac/cluster-scope | kubectl delete -f - --ignore-not-found=true
 fi
+
+# Determine whether deployment is namespace-scoped before deleting runtime resources
+is_namespace_scoped=$(kubectl exec deploy/modelmesh-controller -- printenv NAMESPACE_SCOPE 2> /dev/null || echo "false") || :
 kustomize build default | kubectl delete -f - --ignore-not-found=true
 kustomize build rbac/namespace-scope | kubectl delete -f - --ignore-not-found=true
-kustomize build runtimes ${kustomize_load_restrictor_arg} | kubectl delete -f - --ignore-not-found=true
+if [[ ! "$is_namespace_scoped" == "true" ]]; then
+  echo false
```
ckadner, line 140: `echo false`?

this looks like a dev/test artifact or copy paste error? :-)

should be:

```Bash
   kustomize build runtimes ${kustomize_load_restrictor_arg} | kubectl delete -f - --ignore-not-found=true
```

ckadner (changes requested):
```Bash
  echo false
```
🙃

ckadner (approved):
/lgtm
/approve

ckadner:
Thanks @rafvasq

Issue #424, opened 2023-09-05
Error "setup-envtest: command not found" when running 'make test'


vaibhavjainwiz:
**Describe the bug**

`make test` cmd assumes `setup-envtest` utility would be pre-installed and doesn't install it if its missing.

**To Reproduce**

1. Check if `setup-envtest` utlity is already installed on your system using `which setup-envtest`.
2. If it does then delete the existing `setup-envtest` utlity.
3. Now run `make test`
4. `make test` would print error `/bin/sh: setup-envtest: command not found`

**Expected behavior**

`make test` cmd should install the `setup-envtest` utility if its not already installed.

**Environment (please complete the following information):**

- OS: Mac OS

vaibhavjainwiz:
@ckadner

ckadner:
Thanks @vaibhavjainwiz -- I will take a look

PR #425, opened 2023-09-05
Install setup-envtest utility if not exists

vaibhavjainwiz:
https://github.com/kserve/modelmesh-serving/issues/424


ckadner (changes requested):
@vaibhavjainwiz -- thanks for your PR. It definitely makes sense to auto-install missing dependencies.

Introducing a _local_ `bin` directory is a good best practice to avoid system-wide side effects. However, we already have an established precedent for using the [`GOBIN`](https://github.com/kserve/modelmesh-serving/blob/5dfc278/Makefile#L47-L51) to [install required CLIs](https://github.com/kserve/modelmesh-serving/blob/5dfc278/Makefile#L192). Both for the `Makefile` as well as in [`Dockerfile.develop`](https://github.com/kserve/modelmesh-serving/blob/5dfc278/Dockerfile.develop#L94).

When I tested your changes, I ran into an issue when running `make test` (which creates the local `bin/setup-envtest`) and then run `make run test` inside the developer container:

```
test -s /workspace/bin/setup-envtest || GOBIN=/workspace/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
/workspace/bin/setup-envtest: /workspace/bin/setup-envtest: cannot execute binary file
``` 
Though the developer container already has a "global" `setup-envtest` binary in `GOBIN`.

So, we should either stick with `GOBIN` instead of a `LOCALBIN` or do some extra checking for existing CLIs already available in the current environment.

spolti:
@vaibhavjainwiz is this something that you will continue to work?

PR #428, opened 2023-09-11
Support development version of Kustomize in install script

### LINKED ISSUE #427: Install script not working on kustimize version `(devel)`
**Describe the bug**

The install script just stops without error when it reaches the Kustomize version check.

**To Reproduce**

1. Run `GOBIN=$(pwd)/ GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@latest` to install Kustomize
2. Follow the [quickstart guide](https://github.com/kserve/modelmesh-serving/blob/main/docs/quickstart.md#1-install-modelmesh-serving).

**Expected behavior**

ModelMesh should be installable on Kustomize version `(devel)`.

**Environment (please complete the following information):**

- OS: Ubuntu 22.04

janwytze:
Closes #427 

#### Motivation

Kustomize version `(devel)` is not supported by the install script.

#### Modifications

Changed the regex so `(devel)` is a supported version.

#### Result

Install script now works when Kustomize is installed from https://kubectl.docs.kubernetes.io/installation/kustomize/source/

### DIFF
--- a/scripts/delete.sh
+++ b/scripts/delete.sh
@@ -95,13 +95,7 @@ popd
 
 # Older versions of kustomize have different load restrictor flag formats.
 # Can be removed once Kubeflow installation stops requiring v3.2.
-kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]")
-kustomize_load_restrictor_arg="--load-restrictor LoadRestrictionsNone"
-if [[ -n "$kustomize_version" && "$kustomize_version" < "3.4.0" ]]; then
-    kustomize_load_restrictor_arg="--load_restrictor none"
-elif [[ -n "$kustomize_version" && "$kustomize_version" < "4.0.1" ]]; then
-    kustomize_load_restrictor_arg="--load_restrictor LoadRestrictionsNone"
-fi
+kustomize_load_restrictor_arg=$( kustomize build --help | grep -o -E "\-\-load.restrictor[^,]+" | sed -E "s/(--load.restrictor).+'(.*none)'/\1 \2/I" )
 
 if [[ ! -z $user_ns_array ]]; then
   kustomize build runtimes ${kustomize_load_restrictor_arg} > runtimes.yaml
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -378,13 +378,7 @@ wait_for_pods_ready "-l control-plane=modelmesh-controller"
 
 # Older versions of kustomize have different load restrictor flag formats.
 # Can be removed once Kubeflow installation stops requiring v3.2.
-kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]")
-kustomize_load_restrictor_arg="--load-restrictor LoadRestrictionsNone"
-if [[ -n "$kustomize_version" && "$kustomize_version" < "3.4.0" ]]; then
-    kustomize_load_restrictor_arg="--load_restrictor none"
-elif [[ -n "$kustomize_version" && "$kustomize_version" < "4.0.1" ]]; then
-    kustomize_load_restrictor_arg="--load_restrictor LoadRestrictionsNone"
-fi
+kustomize_load_restrictor_arg=$( kustomize build --help | grep -o -E "\-\-load.restrictor[^,]+" | sed -E "s/(--load.restrictor).+'(.*none)'/\1 \2/I" )
 
 info "Installing ModelMesh Serving built-in runtimes"
 if [[ $namespace_scope_mode == "true" ]]; then
--- a/scripts/setup_user_namespaces.sh
+++ b/scripts/setup_user_namespaces.sh
@@ -72,13 +72,7 @@ if [[ ! -z $user_ns_array ]]; then
 
     # Older versions of kustomize have different load restrictor flag formats.
     # Can be removed once Kubeflow installation stops requiring v3.2.
-    kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]")
-    kustomize_load_restrictor_arg="--load-restrictor LoadRestrictionsNone"
-    if [[ -n "$kustomize_version" && "$kustomize_version" < "3.4.0" ]]; then
-        kustomize_load_restrictor_arg="--load_restrictor none"
-    elif [[ -n "$kustomize_version" && "$kustomize_version" < "4.0.1" ]]; then
-        kustomize_load_restrictor_arg="--load_restrictor LoadRestrictionsNone"
-    fi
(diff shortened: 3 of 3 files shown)


scripts/install.sh
```diff
@@ -378,7 +378,7 @@ wait_for_pods_ready "-l control-plane=modelmesh-controller"
 
 # Older versions of kustomize have different load restrictor flag formats.
 # Can be removed once Kubeflow installation stops requiring v3.2.
-kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]")
+kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]|\(devel\)")
```
rafvasq, line 381: Can you confirm that the logic below for the `kustomize_load_restrictor_arg` continues to work with the change? Seems to me like with kustomize version `(devel)`, I hit `Error: unknown flag: --load_restrictor none`.

rafvasq (changes requested):
Hi @janwytze, thanks for the PR. Along with testing the load restrictor logic below, other scripts would need to include your update as well e.g. [/scripts/setup_user_namespaces.sh](https://github.com/kserve/modelmesh-serving/blob/main/scripts/setup_user_namespaces.sh#L75-L81) and [/scripts/delete.sh](https://github.com/kserve/modelmesh-serving/blob/main/scripts/delete.sh#L98-L104).

scripts/install.sh
janwytze: @rafvasq The output of `kustomize version --short` on my machine is:
```
Flag --short has been deprecated, and will be removed in the future.
{(devel)  unknown   }
```

This causes `$kustomize_load_restrictor_arg` to be set to:
```
--load-restrictor LoadRestrictionsNone
```

I believe this is the correct behavior.

rafvasq: My output is
```
Flag --short has been deprecated, and will be removed in the future.
{(devel)  2023-09-11T18:05:37Z   }
```
and results in `$kustomize_load_restrictor_arg` to be  `--load_restrictor none`.

I'm not sure if it's a variation in how string comparisons are handled across Bash versions, locale settings, or something else, but regardless it I think the logic could be hardened.

rafvasq (changes requested):

ckadner:
I would change the logic away from capturing a semantic version string, to instead capture the fact that kustomize is "old", speak, before 4.x

```Bash
is_kustomize_v3=$(kustomize version --short 2> /dev/null | grep -q -E "3\.[0-9]\.[0-9]" && echo "true" || echo "false")
```

scripts/delete.sh
```diff
@@ -95,7 +95,7 @@ popd
 
 # Older versions of kustomize have different load restrictor flag formats.
 # Can be removed once Kubeflow installation stops requiring v3.2.
-kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]")
+kustomize_version=$(kustomize version --short | grep -o -E "[0-9]\.[0-9]\.[0-9]|\(devel\)")
```
ckadner, line 98:
```shell
is_kustomize_v3=$(kustomize version --short 2> /dev/null | grep -q -E "3\.[0-9]\.[0-9]" && echo "true" || echo "false")
```

scripts/delete.sh
```diff
@@ -95,7 +95,7 @@ popd
...
 kustomize_load_restrictor_arg="--load-restrictor LoadRestrictionsNone"
 if [[ -n "$kustomize_version" && "$kustomize_version" < "3.4.0" ]]; then
```
ckadner, line 100:
```shell
if [[ $is_kustomize_v3 == "true" ]]; then
    kustomize_load_restrictor_arg="--load_restrictor none"
else
    kustomize_load_restrictor_arg="--load_restrictor LoadRestrictionsNone"
fi
```

ckadner (changes requested):
I would change the logic away from capturing a semantic version string (`3.2.0`), to instead capturing if the `kustomize` version is older than 4.x

ckadner:
I removed my earlier review after doing a bit more testing. Will update shortly.

ckadner (changes requested):
Instead of trying to capture the `kustomize` CLI version, we could check for the spelling and parameter value variation of the supported load restrictor flag

```shell
load_restrictor_flag=$( kustomize build --help | grep -o -E "\-\-load.restrictor[^,]+" | sed -E "s/(--load.restrictor).+'(.*none)'/\1 \2/I" )

kustomize build runtimes $load_restrictor_flag
```

Quick test:

```shell

# shell function to set kustomize version, download CLI to user local bin if it doesn't exists
set-kustomize-version() {
  # check and process function parameters
  [ -z "${1}" ] && echo "Usage: '${FUNCNAME[0]} <version>'" && return 1
  [[ ! "$1" =~ ^[0-9]+(\.[0-9]+){2}$ ]] && echo "<version> should be of format 'x.y.z'" && return 1
  VERSION="$1"
  # https://kubectl.docs.kubernetes.io/installation/kustomize/binaries/
  INSTALL_SCRIPT_URL="https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
  # create user-local kustomize install dir 
  INSTALL_DIR=~/bin/kustomize/${VERSION}
  mkdir -p ${INSTALL_DIR}
  # download the requested kustomize CLI, if not done before
  [ -f ${INSTALL_DIR}/kustomize ] || curl -s ${INSTALL_SCRIPT_URL} | bash -s -- ${VERSION} ${INSTALL_DIR}
  # prepend the session PATH with the folder of the desired kustomize CLI version
  export PATH=${INSTALL_DIR}:$PATH
  # report the set kustomize CLI version to the user
  echo -n "Kustomize version: "
  (kustomize version --short 2> /dev/null || kustomize version) | grep -o -E "[0-9]\.[0-9]\.[0-9]"
}

# specify key kustomize versions
kustomize_versions=("3.3.0" "4.0.0" "4.1.0")

# navigate to the "config" folder inside modelmesh-serving repo
cd config

# for each of the kustomize versions, try to build the runtimes with the correct load-restrictor flag
for v in ${kustomize_versions[@]}; do
  set-kustomize-version $v

  load_restrictor_flag=$( kustomize build --help | grep -o -E "\-\-load.restrictor[^,]+" | sed -E "s/(--load.restrictor).+'(.*none)'/\1 \2/I" )
  echo "$load_restrictor_flag"

  kustomize build runtimes $load_restrictor_flag 2>&1 > /dev/null \
    && echo "SUCCESS" \
    || echo "FAILED"

  echo
done

```

Output on MacOS and Linux:

```YAML
Kustomize version: 3.3.0
--load_restrictor none
SUCCESS

Kustomize version: 4.0.0
--load_restrictor LoadRestrictionsNone
SUCCESS

Kustomize version: 4.1.0
--load-restrictor LoadRestrictionsNone
SUCCESS

```

rafvasq (approved):
Tested with versions `3.2.0`, `5.1.1`, and `(devel)`. Thanks @janwytze!

ckadner (approved):
/lgtm

PR #457, opened 2023-11-09
docs: Copy-edit TLS.md

rafvasq:
Copy-edits TLS documentation while testing and fixing commands.


docs/configuration/tls.md
```diff
@@ -23,7 +23,9 @@ distinguished_name = req
 subjectAltName = DNS:modelmesh-serving.${NAMESPACE},DNS:localhost,IP:0.0.0.0
 ```
 
-With the generated key/cert, create a kube secret with contents like:
+`${NAMESPACE}` is the namespace where the ModelMesh Serving Service is deployed.
```
ckadner, line 26: This could go wrong if the variables are not defined, or not replaced by the user. 

I would explicitly set the variables used in this guide like

> First, define the variables that will be used in the shell scripts below. Feel free to change the values of the variables:

```shell
NAMESPACE="modelmesh-serving"  # the controller namespace, where ModelMesh Serving was deployed
SECRET_NAME="modelmesh-certificate"
```

... and make the code snippets work, like


```shell
openssl req -x509 -newkey rsa:4096 -sha256 -days 3560 -nodes \
    -keyout example.key \
    -out example.crt \
    -subj "/CN=${NAMESPACE}" \
    -extensions san \
    -config openssl-san.config
```

... and:

```
kubectl apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  namespace: ${NAMESPACE}
  name: ${SECRET_NAME}
type: kubernetes.io/tls
stringData:
  tls.crt: $(cat example.crt)
  tls.key: $(cat example.key)
  ca.crt: $(cat example.crt)
EOF
```

Or, let `kubectl` create the secret like:

```shell
kubectl create secret generic ${SECRET_NAME} \
        --from-file=tls.key=example.key \
        --from-file=tls.crt=example.crt \
        --dry-run=server \
        -o yaml |
    kubectl -n ${NAMESPACE} apply -f -
```

docs/configuration/tls.md
```diff
@@ -37,13 +39,19 @@ stringData:
   ca.crt: <contents-of-example.crt>
 ```
 
-For basic TLS, only the fields `tls.crt` and `tls.key` are needed in the kube secret. For mutual TLS, add `ca.crt` in the kube secret and set the configuration `tls.clientAuth` to `require` in the ConfigMap `model-serving-config`.
+For basic TLS, only the fields `tls.crt` and `tls.key` are required. For mutual TLS, `ca.crt` should be included and `tls.clientAuth` should be set to `require` in the [`model-serving-config` ConfigMap](./README.md).
+
+You can also create this secret imperatively using:
+
+```
+kubectl create secret tls <secret-name> --cert <cert-file> --key <key-file>
```
ckadner, line 47: similar to above comment, let's make this an executable scriptlet, based on the vars and files defined earlier

```
kubectl create secret tls ${SECRET_NAME} --cert "example.crt" --key "example.key"
```

possibly rename the "example.*" files to something meaningful, like "server.crt" and "server.key" or similar.

docs/configuration/tls.md
```diff
@@ -54,7 +62,7 @@ For basic TLS, only the fields `tls.crt` and `tls.key` are needed in the kube se
           selfSigned: {}
         EOF
 
-3.  Create a `Certificate` CR
+3.  Create a `Certificate` CR:
```
ckadner, line 65: since this is a executable scriptlet, we should define the `${HOSTNAME}` variable as part of the script block (or earlier in this guide) with reasonable default and instruct user to change the value like in the explanation below

docs/configuration/tls.md
```diff
@@ -80,19 +88,17 @@ For basic TLS, only the fields `tls.crt` and `tls.key` are needed in the kube se
...
 
-    If the certificate request is successful, a TLS secret with the PEM-encoded certs will be created as `${SECRET_NAME}`.
+    If the certificate request is successful, a TLS secret with the PEM-encoded certs will be created as `modelmesh-serving-cert`, unless changed above.
 
 4.  Wait for the certificate to be successfully issued
 
-        kubectl get certificate/${SECRET_NAME} --watch
+        kubectl get certificate/modelmesh-serving-cert --watch
 
     Once you see `Ready` as `True`, proceed to the next step.
```
ckadner, line 104: the output below should show the value of `${SECRET_NAME}`  as we define it earlier in this doc

ckadner (changes requested):
Thanks @rafvasq -- your copy-edits look good. I would just make the script blocks more bullet proof :-)

docs/configuration/tls.md
```diff
@@ -4,17 +4,17 @@ TLS can be configured via the `tls.secretName` and `tls.clientAuth` parameters o
...
 
-To create a SAN key/cert for TLS, use command:
+To create a SAN key/cert for TLS, use the following command:
 
 ```shell
 openssl req -x509 -newkey rsa:4096 -sha256 -days 3560 -nodes -keyout example.key -out example.crt -subj '/CN=modelmesh-serving' -extensions san -config openssl-san.config
 ```
 
-Where the contents of `openssl-san.config` look like:
+Where the contents of `openssl-san.config` include:
```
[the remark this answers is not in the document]
rafvasq: That config isn't generated, it's created by the user and passed as into the main command `-config openssl-san.config`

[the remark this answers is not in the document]
ckadner: Oh right!  Thanks Rafael.

ckadner (approved):
Great! Thanks @rafvasq 

/approve
/lgtm

PR #458, opened 2023-11-17
fix: Specify and link python38

rafvasq:
#### Motivation
Build/test failures due to using Python36 in `rest-proxy` and `runtime-adapter` components of ModelMesh.

#### Modifications
Specifying python38 to keep sync between repos.

#### Result
Working builds.


ckadner (approved):
/lgtm
/approve

PR #464, opened 2023-12-02
Update fvt toolchain (#463)

ckadner:
Updating the E2E and performance test toolchain running nightly on IKS


scripts/install.sh
```diff
@@ -338,6 +338,29 @@ if [[ $enable_self_signed_ca == "true" ]]; then
...
+# This happened 7 times as of 2023/12/02:
+#   git log --format=%h --abbrev=7 | grep -E "^[0-9]+$"
+#     7284872
+#     4808804
+#     4746079
+#     0281170
+#     1631706
+#     5293579
+#     4207065
+info "kustomize build default"
```
rafvasq, line 363: Can you elaborate a bit on this comment, did you mean to include it in `scripts/install.sh`? It looks like a reference to [this line](https://github.com/kserve/modelmesh-serving/blob/45cb31b4978ab03ade140cae69da319496effc04/scripts/deploy/iks/deploy-mm-serving.sh#L78).

ckadner: Right, this an error that was tricky to debug. And clearly I have not described it well enough :-)

I added the comment here, since this is the command producing the error. The root cause of the error was in the `sed` command (before/after) found in various places. In case I missed on, or a similar one may get added in the future, I added the comment here to save future developers some time. Instead of a lengthy comment, we could create a separate issue for it and mark it as addressed here.

But I would like to keep a clear path from error to resolution.

```Bash
sed -i.bak 's/newTag:.*$/newTag: '"$GIT_COMMIT_SHORT"'/' config/manager/kustomization.yaml. # before
sed -i.bak 's/newTag:.*$/newTag: "'${GIT_COMMIT_SHORT}'"/' config/manager/kustomization.yaml # after
```

rafvasq: Thanks, I understand now! Personally, I'd prefer capturing that info in an Issue and linking to it with a shorter comment since it's a bit hard to read.

rafvasq (changes requested):

spolti (approved):

scripts/install.sh
ckadner: I will clear it up and clean it up.

ckadner: I moved it to #473

PR #465, opened 2023-12-20
Replace "foo" string with a less generic one

israel-hdez:
If template parsing fails, the "foo" string would appear in pod logs. This should be very hard to happen, since the template is built-in. Changing, simply, to have more meaningful logs.

Fixes opendatahub-io/modelmesh-serving#114

### DIFF
--- a/controllers/config/templating.go
+++ b/controllers/config/templating.go
@@ -49,7 +49,7 @@ func templateSource(r io.Reader, context interface{}) mf.Source {
 	if err != nil {
 		panic(err)
 	}
-	t, err := template.New("foo").Parse(string(b))
+	t, err := template.New("ModelMeshTemplate").Parse(string(b))
 	if err != nil {
 		panic(err)
 	}


spolti:
@israel-hdez it might need a rebase.

controllers/config/templating.go
```diff
@@ -49,7 +49,7 @@ func templateSource(r io.Reader, context interface{}) mf.Source {
 	if err != nil {
 		panic(err)
 	}
-	t, err := template.New("foo").Parse(string(b))
+	t, err := template.New("modelMeshTemplate").Parse(string(b))
```
spolti, line 52: maybe `ModelMeshTemplate`?

israel-hdez: It is fixed.

rafvasq (approved):

israel-hdez:
I think somebody needs to also say `/lgtm`.

rafvasq:
> I think somebody needs to also say `/lgtm`.

I just wanted to give @spolti the final say since he originally requested changes.

israel-hdez:
> > I think somebody needs to also say `/lgtm`.
> 
> I just wanted to give @spolti the final say since he originally requested changes.

He is on vacation right now.
This is not urgent, so we can wait :slightly_smiling_face:

israel-hdez:
Thanks, @rafvasq .

PR #484, opened 2024-01-26
docs: Update model support with Triton's FIL backend

### LINKED ISSUE #185: Can the MLServer runtime be replaced by Triton?
ModelMesh currently uses the MLServer runtime to serve sklearn, xgboost, and lightgbm models. However, it [seems like](https://developer.nvidia.com/blog/real-time-serving-for-xgboost-scikit-learn-randomforest-lightgbm-and-more/) recent versions of Triton now support all of those model types as well.

Does that mean that ModelMesh could support all types of models using a single runtime? And if so, would it be easy to change the adapter to handle this?

rafvasq:
#### Motivation
Triton introduced [support for more model frameworks last year](https://developer.nvidia.com/blog/real-time-serving-for-xgboost-scikit-learn-randomforest-lightgbm-and-more/). 

#### Modifications
- Update the supported models doc
- Doc how to deploy the out-of-the-box example models using Triton which depends on https://github.com/kserve/modelmesh-minio-examples/pull/7. 

#### Result
Closes #185

### DIFF
--- a/docs/model-formats/README.md
+++ b/docs/model-formats/README.md
@@ -12,15 +12,15 @@ By leveraging existing third-party model servers, we support a number of standar
 - [XGBoost](xgboost.md)
 
 | Model Type  | Framework        | Supported via ServingRuntime |
-| ----------- | ---------------- | ---------------------------- |
-| keras       | TensorFlow       | Triton (C++)                 |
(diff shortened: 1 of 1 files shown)


rafvasq:
#485 replaces this PR.

PR #485, opened 2024-01-29
feat: Update Triton model support

### LINKED ISSUE #185: Can the MLServer runtime be replaced by Triton?

rafvasq:
#### Motivation
Triton introduced [support for more model frameworks last year](https://developer.nvidia.com/blog/real-time-serving-for-xgboost-scikit-learn-randomforest-lightgbm-and-more/) and can support xgboost, lightgbm, and more. This PR adds examples and docs to advertise this. 

#### Modifications
- Add newly supported models to Triton runtime config, setting `autoSelect: false`.
- Add an example ISVC config for Triton-served XGBoost model.
- Update example-models doc to reflect example models added in https://github.com/kserve/modelmesh-minio-examples/pull/7
- Update model-formats README to reflect framework support and framework-specific docs to show example ISVC using Triton.
- Add FVTs for lightgbm and xgboost deployment on Triton runtime

#### Result
Closes #185


ckadner (changes requested):
Thank @rafvasq -- I was looking at the failed FVTs which of course are unrelated but it got me thinking about whether we should add a (few) test(s) for the new features, using the new example models?

ckadner (approved):
/lgtm

Thanks @rafvasq

PR #496, opened 2024-03-20
chore: update google.golang.org/protobuf to v1.33.0

israel-hdez:
This is to move away from CVE-20204-24786: infinite loop vulnerability.


rafvasq:
Thanks @israel-hdez,
/lgtm

rafvasq (approved):

PR #498, opened 2024-03-22
chore: Update requirements in custom runtime example

rafvasq:
Updating packages used in recently introduced custom runtime example.

- https://github.com/kserve/modelmesh-serving/security/dependabot/14 
- https://github.com/kserve/modelmesh-serving/security/dependabot/15


ckadner (approved):
/lgtm
/approve

PR #503, opened 2024-04-30
feat: add S390x support in affinity rule

modassarrana89:
Added s390x in node affinity rule as already work is going on to support s390x for tensorflow in runtime-adaptor

### DIFF
--- a/controllers/modelmesh/runtime.go
+++ b/controllers/modelmesh/runtime.go
@@ -341,7 +341,7 @@ func addDomainSocketMount(rts *kserveapi.ServingRuntimeSpec, c *corev1.Container
 func (m *Deployment) addPassThroughPodFieldsToDeployment(deployment *appsv1.Deployment) error {
 	rts := m.SRSpec
 	// these fields map directly to pod spec fields
-	// supported architectures are "amd64", "arm64" and "s390x" 
+	// supported architectures are "amd64", "arm64" and "s390x"
 	deployment.Spec.Template.Spec.NodeSelector = rts.NodeSelector
 	deployment.Spec.Template.Spec.Tolerations = rts.Tolerations
 	archNodeSelector := corev1.NodeSelectorTerm{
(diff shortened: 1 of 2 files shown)


modassarrana89:
@njhill @ckadner @rafvasq Can you guys please review & do the needful

controllers/modelmesh/runtime.go
```diff
@@ -342,16 +342,15 @@ func (m *Deployment) addPassThroughPodFieldsToDeployment(deployment *appsv1.Depl
 	rts := m.SRSpec
 	// these fields map directly to pod spec fields
 	// supported architectures are "amd64" and "arm64", "ppc64le"
-	// and "s390x" are not supported by tensorflow
-	// (https://github.com/kserve/modelmesh-runtime-adapter/pull/38#discussion_r1156749259)
+	// and "s390x"
```
rafvasq, line 345: This comment needs some revision. It seems like the original comment says that both `pp64le` and `s390x` are not supported. The new comment is a bit confusing and looks like they are now both supported by this change. Can you please confirm and revise?

rafvasq (changes requested):
Thanks @modassarrana89, based on the checks it seems like you also need to [lint your changes](https://github.com/kserve/modelmesh-serving/blob/main/docs/developer.md#formatting-and-linting-code) when you get a chance

controllers/modelmesh/runtime.go
modassarrana89: Tensorflow support for s390x and pp64cle is still required & i am currently working on it. Here i have added s390x as part of node affinity. If it is still confusing , do let me know i will update the comment

rafvasq: Thanks for the info, but the comment still needs to be corrected either way. The change you made results in the comment:
```
supported architectures are "amd64" and "arm64", "ppc64le" and "s390x"
```

If all four are supported then it should be:
```
supported architectures are "amd64", "arm64", "ppc64le" and "s390x"
```

But it sounds like they're not, so maybe it's supposed to be something like the following with updated information on the unsupported arch:
```
supported architectures are "amd64", "arm64", and "s390x". "ppc64le" is not yet supported by tensorflow (https://github.com/kserve/modelmesh-runtime-adapter/pull/38#discussion_r1156749259).
```

modassarrana89: Updated comments in latest fix

rafvasq (changes requested):
Sorry for the delay in returning to this @modassarrana89. It looks like you'll have to lint as before.

ckadner:
@modassarrana89 -- I updated your branch and fixed the linting issue

ckadner (approved):
Nice! Thanks @modassarrana89

rafvasq (approved):

PR #515, opened 2024-06-20
Use ubi8 go-toolset:1.21

spolti:
chore:	The ubi8/go-toolset:1.21 was released, we can now revert the develop image
	to this image.


ckadner (approved):
/lgtm

PR #520, opened 2024-07-16
allow external hpa

robinvd:
#### Motivation

The current options for scaling are
- static based on replicas in the servingruntime/config
- dynamic using `hpa` annotation. But this creates a managed hpa that only works with the buildin kubernetes metrics. Adding another hpa will conflict with the already created one.

This PR adds the `external` option where the controller wont set the replicas and wont create an hpa.

#### Modifications

The already existing External AutoscalerClass is now checked for instead of crashing the controller. The behavior is the same as `None`  except it does not set the replicas property.

### DIFF
--- a/config/internal/base/deployment.yaml.tmpl
+++ b/config/internal/base/deployment.yaml.tmpl
@@ -22,7 +22,9 @@ metadata:
     app.kubernetes.io/name: modelmesh-controller
     name: {{.ServiceName}}-{{.Name}}
 spec:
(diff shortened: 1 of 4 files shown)


vinismarques:
Hey @robinvd, have you been using HPA based on GPU load successfully? I would love to see this being integrated.

robinvd:
@vinismarques yes using this setup im running an autoscaling setup based on gpu load.

